PR #116, opened 2014-10-03
Ditch google cdn and add swiping to colorbox theme

t-animal:
I did these mainly for my own installation, but I think these can be useful for the upstream version, too:

bdef57a makes serving of the css and js files by google optional, increasing security and privacy (plus this has benefits for galleries, which are not served by an webserver)
c70900c adds swiping to the colorbox theme (colorbox does not support this out of the box), which is handy if you're watching the images on your mobile device


saimn:
Hi @t-animal 
I have finally merged: I merged the swipe commit manually, and for the cdn setting I choose to not replace the fonts with local ones, it is easier to do and it avoids to increase the size of the zip archive, so it seems a good option.
Thanks for your work on this.

bjoernb:
I would prefer not to load fonts from google but provide them locally.

saimn:
Yes, it would be better. The current patch is a first step. Then there is some work to check all the fonts needed (and include only the fonts actually used to minimize the size), the formats to work on all browsers, etc. Would you like to help on this ?

bjoernb:
Hi Simon,

unfortunately I am very busy right now and do not have the time to dig
into the code. Thanks that you outlined what the remaining tasks
should be.

Greetings,
bjoernb.

Issue #134, opened 2015-01-03
Invalid image kills thumbnail creation


kontza:
Hi,

I was running `sigal build` and noticed that it died on thumbnail creation when the image file was not a valid image file. What had happened, was that the offending image was accidentally replaced with a text file, only the with the name of the original image file.

The failure occurred in _image.py_'s `generate_thumbnail`-function when it ran the line
`img = PILImage.open(source)`

I think that a simple `try-except`around that area would be OK, no need for a fancy rollback, just log the offending file name, or something alone those lines.

saimn:
Hi,
Which version of sigal are you running ? This case should be handled in the latest version at least. 
If you are using the latest and still have the error, can you run in debug mode and paste the full log ?
Thanks

kontza:
Hi,

I have an Ubuntu 14.04.1 LTS system, and sigal 0.9.1 was installed via `sudo pip install sigal`.

Here's the log. When I try-excepted the offending `PILImage.open(source)` and just returned from there, after a image failed -message, the processing went fine.

```
INFO - Reading settings ...
DEBUG - Settings file: sigal.conf.py
DEBUG - Rewrite source : torpantie -> /raid/home/juruotsa/Pictures/torpantie
DEBUG - Rewrite destination : _build -> /raid/home/juruotsa/Pictures/_build
DEBUG - Settings:
{'albums_sort_reverse': False,
 'autorotate_images': True,
 'colorbox_column_size': 4,
 'copy_exif_data': False,
 'destination': u'/raid/home/juruotsa/Pictures/_build',
 'files_to_copy': (),
 'google_analytics': '',
 'ignore_directories': [],
 'ignore_files': [],
 'img_processor': 'ResizeToFit',
 'img_size': (800, 600),
 'index_in_url': False,
 'jpg_options': {'optimize': True, 'progressive': True, 'quality': 85},
 'keep_orig': True,
 'links': '',
 'locale': '',
 'make_thumbs': True,
 'medias_sort_attr': 'filename',
 'medias_sort_reverse': False,
 'orig_dir': 'original',
 'orig_link': False,
 'output_filename': 'index.html',
 'plugin_paths': [],
 'plugins': [],
 'source': u'/raid/home/juruotsa/Pictures/torpantie',
 'theme': 'galleria',
 'thumb_dir': 'thumbnails',
 'thumb_fit': True,
 'thumb_prefix': '',
 'thumb_size': (280, 210),
 'thumb_suffix': '',
[...]

PR #136, opened 2015-01-11
IOError-handler added to gallery.worker to prevent sigal from dying on invalid images.

kontza:
This is related to the issue I filed, https://github.com/saimn/sigal/issues/134, about Sigal dying on invalid image files. E.g. empty ones.

p.s. Sorry about not squashing those commits. I remembered that one only after pushing the first commit.


saimn:
Fixed in 506ad36ad42982fe1a11011dd15c970538155bb4.

PR #143, opened 2015-02-09
Avoid black thumbnails with fade-in videos

jdn06:


saimn:
Hi @jdn06 ,
this could be useful indeed, but Travis is failing and after a quick look it is because the test video's duration is ~2 seconds and you use 2 seconds for the `-ss` parameter ...
I'm not sure what's the best option here:
- extract duration from ffmpeg's output and check if it is > 2s
- add a setting to customize the `ss` vaue.

jdn06:
I chose to add a parameter with a default value of 0. As everybody doesn't have the same kind of fade-in, it seems more useful than a check of the length of the video.

saimn:
Thanks, but the tests still fails, could you check why ?

saimn:
Great, thanks !

PR #148, opened 2015-03-02
Watermark Plugin

abdul:
This plugin allows adding watermark to images.


sigal/plugins/watermark.py
```diff
@@ -0,0 +1,77 @@
+# -*- coding: utf-8 -*-
+
+"""Plugin which adds a watermark to the image.
+
+Settings:
+
+- ``watermark``: path to the watermark image.
+- ``watermark_position``: the watermark position (scale or tile)
```
saimn: From the code it seems that it is also possible to specify the position with a tuple which is given to http://pillow.readthedocs.org/reference/Image.html#PIL.Image.Image.paste
Could you also document this behavior ?

sigal/plugins/watermark.py
```diff
@@ -0,0 +1,77 @@
...
+
+def add_watermark(img, settings=None):
+    logger.debug('Adding watermark to %r', img)
+    mark = Image.open(settings['watermark'])
+    position = 'scale'
+    opacity = 1
+    if settings['watermark_position']:
+        position = settings['watermark_position']
+    if settings['watermark_opacity']:
+        opacity = settings["watermark_opacity"]
```
saimn: You could use `settings("watermark_opacity", 1)` to simplify the code, the same for `watermark_position`.

sigal/plugins/watermark.py
```diff
@@ -0,0 +1,77 @@
...
+        mark = mark.resize((w, h))
+        layer.paste(mark, ((im.size[0] - w) / 2, (im.size[1] - h) / 2))
+    else:
+        layer.paste(mark, position)
+    # composite the watermark with the layer
+    return Image.composite(layer, im, layer)
+
+def add_watermark(img, settings=None):
+    logger.debug('Adding watermark to %r', img)
+    mark = Image.open(settings['watermark'])
```
saimn: The path here is relative to the cwd. To allow running sigal from another directory, and be consistent with other path settings, could you add the `watermark` setting here https://github.com/saimn/sigal/blob/master/sigal/settings.py#L126 to make it relative to the settings file ?

saimn:
Looks good (with some minor comments), thanks for the contribution !

abdul:
I would make these changes, and update the PR. Thanks for the feedback.

abdul:
@saimn I have update the code, please review. Thank you.

saimn:
Thanks !

abdul:
Thank you.

On Friday, March 6, 2015, Simon Conseil <notifications@github.com> wrote:

> Thanks !
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/saimn/sigal/pull/148#issuecomment-77564692>.
>


-- 
Sent from Gmail Mobile

PR #151, opened 2015-03-16
Honor 'use_orig' for videos

jasuarez:
When 'use_orig' is True, copy the original video files (or symlink them,
depending if 'orig_link' is True) to the destination, and use them as they are,
without processing.


jasuarez:
Withdrawing the PR until I fix the tests.

PR #152, opened 2015-03-16
Colorbox improvements

jasuarez:
A couple of minor improvements for colorbox theme:
- Show video title instead of video filename when opening a video.
- As it is done for albums, lets write below each media its title.

### DIFF
--- a/sigal/themes/colorbox/static/css/style.css
+++ b/sigal/themes/colorbox/static/css/style.css
@@ -48,7 +48,7 @@ header h1 a {
   box-shadow: 0 0 5px #818181;
 }
 
-.album_title {
+.album_title, .media_title {
   display: block;
 }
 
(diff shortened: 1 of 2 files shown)


saimn:
I have tested this, the idea is fine but when some images have a title and other don't this introduce a gap between images. To solve this one could you the height of images, but I'm not sure of the best way to do this. I will look if I find some time.

jasuarez:
I've managed to reproduce the same issue, but in a different way: having different lengths in the title can also raise the problem.

jasuarez:
Rebased PR against latest master

saimn:
Hi @jasuarez ,

- About the title instead of filename, I think it was changed in b88d10a15e8533fd9032920636f28be90bca909f
- For the title below media, do you see a solution to avoid breaking the layout ? Thanks to #164, the size of the images is now available in the images properties.

PR #153, opened 2015-03-16
Honor `use_orig` with videos

jasuarez:
So far `use_orig` is totally ignored for the case of videos.

But giving that lot of times we have a big amount of videos that we don't want to process because it is expensive, or they are already web-friendly.

This set of commits makes `use_orig` to be respected in case of videos.

I was doubting between following this, or adding a new `video_use_orig` (and `video_orig_link`). But at the end I decided to use the same property for both cases.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -44,7 +44,7 @@ from .compat import PY2, UnicodeMixin, strxfrm, url_quote, text_type
 from .image import process_image, get_exif_tags, get_exif_data
 from .settings import get_thumb
 from .utils import (Devnull, copy, check_or_create_dir, url_from_path,
-                    read_markdown, cached_property)
+                    read_markdown, cached_property, is_valid_html5_video)
 from .video import process_video
 from .writer import Writer
 
@@ -176,11 +176,12 @@ class Video(Media):
 
     def __init__(self, filename, path, settings):
         super(Video, self).__init__(filename, path, settings)
-        base = splitext(filename)[0]
+        (base, ext) = splitext(filename)
         self.date = None
         self.src_filename = filename
-        self.filename = self.url = base + '.webm'
-        self.dst_path = join(settings['destination'], path, base + '.webm')
+        if not settings['use_orig'] or not is_valid_html5_video(ext):
+            self.filename = self.url = base + '.webm'
+            self.dst_path = join(settings['destination'], path, base + '.webm')
 
 
 class Album(UnicodeMixin):
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -44,7 +44,8 @@ from .compat import PY2, UnicodeMixin, strxfrm, url_quote, text_type
 from .image import process_image, get_exif_tags, get_exif_data
 from .settings import get_thumb
 from .utils import (Devnull, copy, check_or_create_dir, url_from_path,
-                    read_markdown, cached_property, is_valid_html5_video)
+                    read_markdown, cached_property, is_valid_html5_video,
+                    get_mime)
 from .video import process_video
 from .writer import Writer
 
@@ -181,7 +182,10 @@ class Video(Media):
         self.src_filename = filename
         if not settings['use_orig'] or not is_valid_html5_video(ext):
             self.filename = self.url = base + '.webm'
+            self.mime = get_mime('.webm')
             self.dst_path = join(settings['destination'], path, base + '.webm')
+        else:
+            self.mime = get_mime(ext)
 
 
 class Album(UnicodeMixin):
--- a/sigal/themes/colorbox/templates/index.html
+++ b/sigal/themes/colorbox/templates/index.html
@@ -84,7 +84,7 @@
         <div style='display:none'>
           <div id="{{ mhash }}">
             <video controls>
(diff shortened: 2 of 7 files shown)


sigal/video.py
```diff
@@ -75,21 +75,29 @@ def video_size(source):
...
     logger = logging.getLogger(__name__)
 
+    if settings['use_orig']:
+        # Keep original filename
+        filename = os.path.split(source)[1]
+        outpath = os.path.split(outname)[0]
+        outname = os.path.join(outpath, filename)
+        utils.copy(source, outname, symlink=settings['orig_link'])
+        return outname
+
```
saimn: This should probably go at the beginning of `process_video`, as it would avoid to modify the path here and avoid to return a modified filename.

sigal/themes/colorbox/templates/index.html
```diff
@@ -120,7 +120,7 @@
               <div style='display:none'>
                 <div id="{{ media.filename|replace('.', '')|replace(' ', '') }}">
                   <video controls>
-                  <source src='{{ media.filename }}' type='video/webm' />
+                  <source src='{{ media.filename }}' {% if not settings.use_orig %}type='video/webm'{% endif %}/>
```
saimn: What happens when the type is not specified ?

tests/test_gallery.py
```diff
@@ -114,6 +114,7 @@ def test_image(settings, tmpdir):
 
 def test_video(settings, tmpdir):
     settings['destination'] = str(tmpdir)
+    settings['use_orig'] = False
```
saimn: This surprised me a bit, `use_orig` should be False by default, but it seems that the settings dict is modified by `test_media_orig`, which is bad. I will try to fix this (unless you want to take a look :-)).

sigal/gallery.py
```diff
@@ -179,8 +179,9 @@ def __init__(self, filename, path, settings):
         base = splitext(filename)[0]
         self.date = None
         self.src_filename = filename
-        self.filename = self.url = base + '.webm'
-        self.dst_path = join(settings['destination'], path, base + '.webm')
+        if not settings['use_orig']:
+            self.filename = self.url = base + '.webm'
+            self.dst_path = join(settings['destination'], path, base + '.webm')
```
saimn: Maybe add a test on the extension to make sure that it is supported by the video tag (webm, mp4, ogv) ?

saimn:
Thanks for this PR, it is indeed something that was missing (as other feature like mp4/multiple format support).

sigal/video.py
jasuarez: Correct. I added it in _generate_video_ because the same is done in _generate_image_. But I'll move it to _process_video_.

tests/test_gallery.py
jasuarez: Yes, I was quite puzzled about this issue.

sigal/themes/colorbox/templates/index.html
jasuarez: Nothing :)

I mean, I've been testing in Firefox and Epiphany, and didn't perceive anything weird: videos were played normally, without any kind of warning.

sigal/gallery.py
jasuarez: Good point.

Actually, I wonder how it should behave when it finds out a video not supported, like an avi file. So far it just, it just copies the file in the output. But when opening with a browser, the video can't be played.

I see here two options:
- Copy the file anyway, but provides a warning message in the output telling the file is not supported.
- Adding a new option, like `convert_unsupported` or similar, that performs a conversion on these cases. So the avi file would be converted in webm.

tests/test_gallery.py
saimn: Fixed with 94c98588b8d0009cf11e460c85c9a96a51104cb2

sigal/gallery.py
saimn: Sorry I forgot to answer. I think that unsupported files should be converted without taking care of the `use_orig` setting, and without adding a new setting. Maybe just make it clear in the doc that `use_orig` applies only to supported files.

sigal/themes/colorbox/templates/index.html
saimn: I would be easy to add the type as an attribute of the Video object, based on the file extension, so it is probably better to do it just in case.

saimn:
Hi @jasuarez ,
I'm thinking about releasing a new version, any chance to get this in it ?

jasuarez:
i'm bit busy these weeks, but I'll try to address your comments this weekend

jasuarez:
@saimn Re-created the PR addressing your comments, plus rebasing against latest master.

saimn:
Thanks @jasuarez ! Could you just revert the change in test_video as it should no more be necessary ?

saimn:
Also please add yourself to AUTHORS ;)

jasuarez:
@saimn I need to push the change in `test_video` because otherwise I'm breaking the tests.

saimn:
Thanks, merged in 2f90da571361d588eb6b502e5aff37fd7d6fdc53

> I need to push the change in test_video because otherwise I'm breaking the tests.
No, before all tests shared the settings dict, but since a recent change on master there is a settings doct for each test.

PR #159, opened 2015-05-24
added support for mp4, manually tested, mp4 files transcoded and link…

MPavleski:
…s correctly populated in html


saimn:
Thanks for this PR ! Could you add the `video_format` setting in the settings template (` sigal/templates/sigal.conf.py`) and add yourself to the AUTHORS file ? It would also be good to have a test for this, or to parametrize the tests in `tests/test_video.py` to run for both webm and mp4 (using http://pytest.org/latest/parametrize.html#parametrize).

MPavleski:
Thank you for the good work saimn. Hope the PR is ok now.

MPavleski:
I see the travis is failing (on my windows machine with ffmpeg it passes). I do not know it that ffmpeg supports mp4..

saimn:
I have tested your pr and it doesn't work with your settings (`-crf 23`). I get this error (with ffmpeg version 2.6.3):
```
The encoder 'aac' is experimental but experimental codecs are not enabled, add '-strict -2' if you want to use it.
```
If I add `-strict -2`, then it is ok. Another option could be to use another audio encoder, `-acodec libmp3lame` for instance.I don't know what is the best default settings ?

MPavleski:
I researched a bit and the safest option would be to go with -strict -2 which will use the internal encoder. Default installations usually have another encoder as well libvo_aacenc, but doc. says it is performing worse that the built-in one (the experimental).

saimn:
Ok thanks Miroslav, can you add this to the default options ?

MPavleski:
Hey I've added this while back .. is it still not good ?

saimn:
Hi Miroslav,
Github doesn't notify when you just add a commit, so I didn't see it ;)
Thanks for your work on this !

iemejia:
Hi, sorry to comment in this already closed PR. This change is worth a new public version, it has been some time since the last release, and I think users can benefit of this feature (also me, since I am one of the users of the PIP based release (since I never arrived to make it work from the source code).

saimn:
Sorry for the late answer; I plan to make a new release since several weeks but time flies ... It will come sooner or later.

PR #160, opened 2015-05-24
Fix for the relative check in Windows which fails due invoking relati…

MPavleski:
…ve path checking for directories on different drives

### DIFF
--- a/sigal/__init__.py
+++ b/sigal/__init__.py
@@ -117,11 +117,14 @@ def build(source, destination, debug, verbose, force, config, theme, title,
         logger.error("Input directory not found: %s", settings['source'])
         sys.exit(1)
 
+    # on windows os.path.relpath raises a ValueError if the two paths are on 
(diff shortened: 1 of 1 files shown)


sigal/__init__.py
```diff
@@ -117,8 +117,14 @@ def build(source, destination, debug, verbose, force, config, theme, title,
         logger.error("Input directory not found: %s", settings['source'])
         sys.exit(1)
 
-    if not os.path.relpath(settings['destination'],
-                           settings['source']).startswith('..'):
+    relative_check = True
+    try:
+        relative_check =  os.path.relpath(settings['destination'],
+                           settings['source']).startswith('..')
+    except:
```
saimn: Could you specify here the type of exception that must be catched and add a comment ?

saimn:
Thanks !

MPavleski:
Thank you for the excellent tool.

PR #164, opened 2015-06-12
add size property into Image object containing width and height of each image (thumbnail, big).

franek:
Can be useful for some themes to disinguish orientation of the image.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -171,21 +171,11 @@ class Image(Media):
 
     @cached_property
     def size(self):
-        try:
-            return get_size(self.dst_path)
-        except (IOError, IndexError, TypeError, AttributeError):
-            self.logger.warning(u'Could not read size %s',
-                                self.dst_path)
-            return None
+        return get_size(self.dst_path)
 
     @cached_property
     def thumb_size(self):
-        try:
-            return get_size(self.thumb_path)
-        except (IOError, IndexError, TypeError, AttributeError):
-            self.logger.warning(u'Could not read size %s',
-                                self.thumb_path)
-            return None            
+        return get_size(self.thumb_path)
 
 class Video(Media):
     """Gather all informations on a video file."""
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -41,7 +41,7 @@ from PIL import Image as PILImage
 
 from . import image, video, signals
 from .compat import PY2, UnicodeMixin, strxfrm, url_quote, text_type
-from .image import process_image, get_exif_tags, get_exif_data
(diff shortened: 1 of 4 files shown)


saimn:
It would be a good thing to add this, thanks for proposing it. However I would like a different api: `get_size` should work on only one file (not specific case for the thumbnail, which is just another file), and there should be two properties, `Image.size` and `Image.thumb_size` for the thumbnail.
Actually I'm not sure if the `get_size` function is needed, but one advantage of putting this in a function should be to have a unit test, it would great if you can add one ;).

sigal/gallery.py
```diff
@@ -371,7 +379,7 @@ def thumbnail(self):
                     except:
                         self.logger.error("Failed to open %s", f.src_path)
                     else:
-                        if im.size[0] > im.size[1]:
+                        if im.size[0] > im.size[1]:
```
saimn: In this block of code there is a call to `PILImage.open` which could be simplified by using the new property, if you want to change that.

franek: I have tried but I am a python "hacker" (bidouilleur) and did not succeed (got some errors with tests). Sorry about that.

sigal/image.py
```diff
@@ -160,6 +161,20 @@ def process_image(filepath, outpath, settings):
     return Status.SUCCESS
 
 
+def get_size(file_path):
+    """Return image size (width and height)."""
+    logger = logging.getLogger(__name__)
+    try:
+        im = PILImage.open(file_path)
+    except:
+        logger.error("Failed to open %s", file_path)
```
saimn: With this try/except, the try/except in the properties is useless and can be removed. Also the error message could be more specific, like above (`Could not read size ...`), and specifying the list of exceptions (`IOError, IndexError, TypeError, AttributeError` should be fine) is better practise.

tests/test_image.py
```diff
@@ -115,3 +115,13 @@ def test_exif_gps(tmpdir):
...
+
+def test_get_size(tmpdir):
+    """Test reading out image size"""
+
+    test_image = 'flickr_jerquiaga_2394751088_cc-by-nc.jpg'
+    src_file = os.path.join(CURRENT_DIR, 'sample', 'pictures', 'dir1', 'test1',
+                            test_image)
+
+    result = get_size(src_file)
+    assert result == {'height': 800, 'width': 600}
```
saimn: Please add a test for a missing file to test the except clause.

sigal/gallery.py
saimn: Ok, no problem, I will take a look.

saimn:
Thanks François !

PR #165, opened 2015-07-23
add support for piwik

franek:
Sigal only provide google analytics tracker. The main purpose of this PR is to add piwik support.
User need to configure two settings : 
* tracker_url : url of the piwik instance (http://stats.domain.com or http://www.domain.com/piwik)
* site_id : unique identifier of the tracking site


saimn:
Useful, thanks !

PR #185, opened 2015-12-15
Add GIF support — Fixes #140

### LINKED ISSUE #140: Support for GIF images
I have set up a gallery with sigal and would much appreciate if the .gif extension could be added to the list of formats recognized as an image. 

It seams the toolchain already works quite well with gifs, if I just rename the files to .jpg they will get thumbnails etc, but it complains about not being able to read exif data. 

Gifs can be quite useful for animated photobursts for example.

Thanks for a great gallery tool btw!

Natim:


saimn:
Great to see some work on this, thanks Rémy !
Last time I looked at GIF support, I was not sure how to handle the resize operations. You solved the first one, using a copy for the normal size image, which is fine, but there is also the generated thumbnail. Is it working ? 
It would be nice to have a test image in `tests/sample/pictures/` to check if it works. 
Also please add yourself to AUTHORS :)

saimn:
Ref #140

Natim:
As for video I didn't created an animated thumbnail.
Also it looks like GIF are already optimised (and are probably shipped in a size lower than the one we want to resize them in).

Can you help me to list the tests you want to add?

 - One to see that gif aren't ignored
 - One to see that gif file are thumbnailed
 - One to see that gif image are copied and not resized.

saimn:
Your list seems good. There is a test that simply ensures that the test gallery can be build, so if there is a GIF it is already a start. Then I just want to be sure that we can generate a thumbnail for a GIF file. The thumbnail should probably be a PNG file ?

Natim:
Apparently for now they are gif files as well but not animated ones.

saimn:
Ok, then it's fine. I was not sure that Pillow is able to resize a GIF.

saimn:
Hi @Natim, any update on this ?

Natim:
I tried but I didn't really find out where to put my file.

I added tests/sample/pictures/dir1/test1/50a1d0bc-763d-457e-b634-c87f16a64270.gif but only `test_zipped_correctly` is failing did I miss something?

Natim:
- Adding a PNG or GIF file in ``tests/sample/pictures/dir1/test2`` didn't seems to change anything
- Adding a PNG or GIF file in ``tests/sample/pictures/dir1/test1/`` is just making test_zipped_correctly fails
- Adding a PNG or GIF file in ``tests/sample/pictures/dir2/`` doesn't change anything
- Adding a PNG or GIF file in ``tests/sample/pictures/empty`` actually make ``test_empty_dirs`` fails which is a good news :+1: 
- Adding a PNG or GIF file in ``tests/sample/pictures/accentué`` doesn't change anything.

When looking at the test_galery you are mentionning earlier I think changing the directory file list should make tests fails because the hardcoded REF wasn't modified. Can you help me with this ?

Natim:
Apparently the tests in ``test_gallery`` are just verifying that when we create an Album the parameters are correctly exposed.

    a = Album('dir1/test1', settings, album['subdirs'], album['medias'], gal)
    assert list(im.filename for im in a.images) == album['medias']

Is that what we want? Don't we want to see that the directory is correctly parsed?

Natim:
Ok so my last commit seems to do what you wanted as a first test. I am going to write one for the thumbnail as well.

saimn:
Yeah the tests are not complete :(
`test_album_medias` just tests that the `Album` objet is correctly created but it doesn't parse the directory, the list of files is given in input. So actually the REF dict is not used to test that all the files have been parsed. 
And `test_gallery` just tests that the gallery builds without errors. It would be a good idea to improve this test to check with REF that all files have been parsed, but you can leave that for now. Having at least one GIF file processed in these tests is enough for now.

saimn:
Please add yourself to AUTHORS :)

Natim:
Here you go, fully tested :+1:

saimn:
Thanks for adding GIF in the image tests, though it can be made simpler to avoid code duplication. As the tests you added seems exactly the same as `test_generate_thumbnail` and `test_generate_image_passthrough`, you could use pytest's parametrize feature:
```
@pytest.mark.parametrize("image", [TEST_IMAGE, TEST_GIF_IMAGE])
def test_generate_image_passthrough(tmpdir, image):
    dstfile = str(tmpdir.join(image))
    ...
```

Natim:
Yes sure let's do that :+1:

Natim:
That's not awful but not really pretty either...

saimn:
Seems good, many thanks Rémy !

almet:
You both rock :-)

Issue #172, opened 2015-10-04
EXIF data reading is broken with Pillow 3.0


saimn:
It seems that Pillow 3.0 changes the format of EXIF data.
http://pillow.readthedocs.org/en/latest/releasenotes/3.0.0.html

With Pillow 3.0:
```
 'ExposureTime': (0.00100603,),
 'FNumber': (3.9,),
 'FileSource': (b'\x03',),
 'Flash': (16,),
 'FlashPixVersion': (b'0100',),
 'FocalLength': (11.5,),
 'FocalLengthIn35mmFilm': (76,),
 'GainControl': (0,),
 'ISOSpeedRatings': (50,),
```

With Pillow 2.9.0:
```
 'ExposureTime': (100603, 100000000),
 'FNumber': (39, 10),
 'FileSource': 3,
 'Flash': 16,
 'FlashPixVersion': b'0100',
 'FocalLength': (115, 10),
 'FocalLengthIn35mmFilm': 76,
 'GainControl': 0,
 'ISOSpeedRatings': 50,
```

saimn:
Fixing this is straightforward, but there is a more serious issue with the GPSInfo which is broken in Pillow 3.0. Let's hope it will be fixed in the next bugfix release (https://github.com/python-pillow/Pillow/issues/1477)

willmcgugan:
What's the verdict on this one? Is the format change intentional or a bug? A tuple of one value is a little odd.

I'd recommend keeping fractions as list of two values, rather than converting to a float. Photographers will want to see a photo had a shutter speed of 1/640, not 0.0015625. Working backwards would be a pain there.

Could you keep the `_getexif` method returning the same format as it did pre-3.0.0, for the sake of backwards compatibility. And add a `getexif` for any format changes?

saimn:
I guess your message should be addressed to Pillow developers, there is nothing I can do about this on Sigal's side.

willmcgugan:
My bad. I followed a link. Thought I was in the Pillow project.

saxenanurag:
Any updates on this? I have temporarily switched to 2.9 to get GPSData from image. Thanks.

saimn:
@saxenanurag No, it seems that nothing happened on Pillow side since a few weeks ...

wiredfool:
For everyone who's interested in the exif handling in Pillow, Please test the new 3.1.0-rc1 release soon and give feedback.

saimn:
@wiredfool Seems good, thanks !

[...]

PR #187, opened 2015-12-17
better exposure time fraction

dsiroky:


saimn:
It's fine, but this EXIF reading stuff is currently blocked by Pillow's breaking changes in 3.0 which is still not fixed (see #172):
- You changed only the code specific to Pillow 2.9.0. With Pillow 3.0 it is not possible to get the numerator and denominator. I think one proposed fix on Pillow side is to use fractions.
- Tests need to be updated, but Pillow 2.9.0 and 3.0 will not give the same result. 
So it is probably better to wait for Pillow to see if they will finally do something (but there is activity since ~1 month ...).

wiredfool:
Please test the new Pillow 3.1.0-rc1 release and give feedback.

saimn:
Now that Pillow 3.1.0 is out, I got back on this and merged manually in https://github.com/saimn/sigal/commit/6e279bbff44d360d9db98a07848d9a67ceee8ad5
Thanks !

Issue #121, opened 2014-10-22
Sort albums by 'Date'


jrstarke:
I really like this project, and I think there are only a few small things to get it to a point I love it.  

One of the things I'd like to be able to do is order my albums by reverse chronological (most recent at the top) so that when a user comes to my page, they see my most recent photos first. 

I can think of two ways of achieving this, but I'd like to get your thoughts. 

1. Add a `Date` property to the index.md file in the album, and add a configuration flag to sort by `date` to the config. When sorting the albums, if the `date` flag is set, order by this date, or alphabetic if none is available.
2. For each album, get the most recent exif from the medias, and then sort the albums based on their most recent media.

For complexity reasons, I'm more inclined to 1, but I figured it might be good to get some feedback before attempting to build something.

saimn:
Fixed with #192

PR #192, opened 2016-01-18
Implement album sorting

trapperhoney:
This adds an 'albums_sort_attr' configuration modeled off of the
'medias_sort_attr'.  It defaults to the existing behavior of sorting
sub-albums by directory name.  It allows sort by any metadata attribute
stored in the album metadata file.

The existing album sort test wasn't really verifying album sorting
behavior, so I just re-wrote the whole thing to include tests for the
old behavior and new.

This should resolve issue #121


saimn:
Looks good, thanks @trapperhoney !
Could you remove the `.index.md.swp` file and add yourself to the AUTHORS file ?

trapperhoney:
Woops, sorry I missed the tmp file :\  I'd rather not include my name in the distributed source.  Is the github trail not good enough?

saimn:
It can be your github pseudonym if you prefer, no problem with this.

saimn:
Thanks !

trapperhoney:
Thank you!

tohojo:
How is this supposed to work? I tried adding an 'Order' field to index.md of each of my subdirectories and setting that as album_sort_attr. However, I just get an AttributeError when trying to build the gallery.

I ended up adding this to _get_metadata() of the Album class, but that's obviously not a workable solution to have to do that for all the interesting fields:

        try:
            self.order = self.meta['order'][0]
        except KeyError:
            self.order = self.settings.get('order')

saimn:
@tohojo - Indeed, you're right, it works only with attributes of the Album objects ... Copying all the `meta` items as object attributes seems not a good option, so maybe we could allow some special syntax for albums_sort_attr, something like ` 'albums_sort_attr': ('meta', 'Order') ` ?

tohojo:
Ah, good, so it wasn't just me who was missing something! :)

Yeah, having a way to specify meta fields would be good (thought that was what it was meant for). Maybe just falling back to looking in the metadata object if the attribute is not found on the Album object itself? That would avoid introducing new syntax... Not sure if there are field names that would clash in surprising ways, though?

saimn:
> Maybe just falling back to looking in the metadata object if the attribute is not found on the Album object itself? That would avoid introducing new syntax..

Yep, that's another possibility. Maybe both could be allowed to resolve potential conflicts.

tohojo:
Ah yes, that's a possibility. Can take a shot at a patch tomorrow. Any preferences on the syntax? Was thinking using 'meta.fieldname' could work?

trapperhoney:
For what it's worth, the same (lack of metadata support) semantics apply to the sorting behavior of individual items and the items_sort_attr configuration option.

saimn:
@tohojo - Great ! No string preference, `meta.fieldname` can work too.
@trapperhoney - Yes, good point. The media objects have more useful attributes, but their metadata is not accessible for the sort.

tohojo:
Created #202 :)

PR #195, opened 2016-02-25
Enable autoescape in jinja templates

dairiki:
Special HTML characters in album and media metatdata were not being escaped in the generated HTML pages. So if a album title contained a ``<``, for example, there was trouble.

This pull request turns on jinja2’s autoescape feature, which solves the problem.


saimn:
Thanks !

PR #199, opened 2016-04-07
Empty ISO speed ratings

thekad:
I found a couple of images where the ISOSpeedRatings value is an empty tuple. This makes sigal sad. Added a test and also fixed the issue so the test passes now.

PS CONTRIBUTING.rst mentions "squash" commits, since github now supports squashing pull requests, is this still necessary? https://github.com/blog/2141-squash-your-commits


saimn:
Thanks !
From what I saw the squashing feature of github will not make a merge commit, which I think is useful to better track merges of pull requests.

PR #202, opened 2016-04-27
Allow sorting on metadata keys.

tohojo:
This makes it possible to specify metadata keys for the albums_sort_attr and medias_sort_attr configuration options. This is done by prefixing the key with 'meta.', so e.g. albums_sort_attr = 'meta.order' will sort albums on the 'order' metadata key (collected from the .md files in each directory).

Currently, there's no fallback so the metadata key has to exist for all albums/images and wasn't sure how to add tests (there doesn't seem to be any metadata handling in the test cases at all currently?). If you can provide some feedback on how best to do this I can update the pull request :)


saimn:
Thanks for working on this @tohojo !
There are some tests here: https://github.com/saimn/sigal/blob/master/tests/test_gallery.py#L167 (`medias_sort_attr` is tested, but not `albums_sort_attr` it seems).
It would be nice also if it doesn't fail if some md file doesn't have the meta. Maybe with something like
`.meta.get(meta_key, [''])[0]`, which will make a long and ugly line, but ... And it will put the items without the meta key at the beginning which is not ideal.

tohojo:
Okay, added some tests. Adding an index.md file to the 'test2' directory in the test tree broke another test, so changed the behaviour of the read_markdown() function.

This is obviously a genuine change of behaviour, but it made sense to me: Why would simply having an index.md file with no title defined not still fall back to the directory name? However, if you prefer I do something else to fix the tests, let me know and I can change it :)

saimn:
> Adding an index.md file to the 'test2' directory in the test tree broke another test, so changed the behaviour of the read_markdown() function.

Yes indeed, it was a bug so your fix is fine.

tests/test_gallery.py
```diff
@@ -191,6 +191,18 @@ def test_albums_sort(settings):
     a.sort_subdirs('title')
     assert [im.title for im in a.albums] == list(reversed(titles))
 
+    orders = [d.meta['order'] for d in a.albums]
```
saimn: This list could be hard-coded as it is the reference to do the test. It would avoid to rely on the metadata parsing code which is used later in the sort.

tests/test_gallery.py
```diff
@@ -191,6 +191,18 @@ def test_albums_sort(settings):
...
+    orders.sort()
+    settings['albums_sort_reverse'] = False
+    a = Album('dir1', settings, album['subdirs'], album['medias'], gal)
+    a.sort_subdirs('meta.order')
+    assert [d.meta['order'] for d in a.albums] == orders
```
saimn: `d.meta['order'][0]` is maybe safer, I don't know if and how py.test tests recursively list of lists.

tests/test_gallery.py
tohojo: That is true; can do.

tests/test_gallery.py
tohojo: I think the comparison is fine; Python considers two lists to be equal if all their elements are. But if I'm hardcoding the list anyway, it's probably better to change it :)

saimn:
If you can handle the case when one of the md files does not have the meta used for the sort, then we are good. Also please add yourself to the AUTHORS files ;). Thanks !

tohojo:
Ah yes, totally forgot about that. Will do!

tohojo:
Updated... Should be good to go now :)

saimn:
Great, thanks @tohojo !

PR #203, opened 2016-04-27
Add 'title' command.

tohojo:
I found myself wanting a quick way to add simple titles to lots of images. Rather than writing a separate script I thought that just adding a new command to sigal itself might be easier. Do you think this is worth including in upstream sigal? In that case I can add test cases and documentation for it, but wanted to see if you'd want to take it at all first...

### DIFF
--- a/sigal/__init__.py
+++ b/sigal/__init__.py
@@ -223,3 +223,23 @@ def serve(destination, port, config):
         httpd.serve_forever()
     except KeyboardInterrupt:
         print('\nAll done!')
+
+@main.command()
+@argument('image')
+@argument('title')
+@option('-f', '--force', default=False, is_flag=True,
+        help='Overwrite existing .md file')
+def title(image, title, force=False):
+    "Write image title to corresponding .md file"
+
+    if not os.path.exists(image):
+        sys.stderr.write("The image {} does not exist.\n".format(image))
+        sys.exit(1)
+    descfile = os.path.splitext(image)[0] + '.md'
+    if os.path.exists(descfile) and not force:
+        sys.stderr.write("Description file for {} already exists. "
+                         "Use --force to overwrite it.\n".format(image))
+        sys.exit(2)
+
+    with open(descfile, "w") as fp:
+        fp.write("Title: {}\n".format(title))
--- a/sigal/__init__.py
+++ b/sigal/__init__.py
@@ -225,21 +225,38 @@ def serve(destination, port, config):
         print('\nAll done!')
 
 @main.command()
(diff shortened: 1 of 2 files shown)


saimn:
It can be useful so +1 but maybe a bit more generic, to allow writing several key/value pairs at once ? Then we also need a more generic name ;)

sigal/__init__.py
```diff
@@ -223,3 +223,23 @@ def serve(destination, port, config):
         httpd.serve_forever()
     except KeyboardInterrupt:
         print('\nAll done!')
+
+@main.command()
+@argument('image')
+@argument('title')
+@option('-f', '--force', default=False, is_flag=True,
```
saimn: I would prefer `overwrite` to `force`, more explicit on what it does.

saimn:
It could be useful for albums as well.

tohojo:
Something like this?

sigal/__init__.py
```diff
@@ -223,3 +223,37 @@ def serve(destination, port, config):
...
     except KeyboardInterrupt:
         print('\nAll done!')
+
+@main.command()
+@argument('target')
+@argument('keys', nargs=-1)
+@option('-o', '--overwrite', default=False, is_flag=True,
+        help='Overwrite existing .md file')
+def set_meta(target, keys, overwrite=False):
+    """Write image title to corresponding .md file.
```
saimn: Docstring needs to be updated, mention that target can be a directory or a media file.

sigal/__init__.py
```diff
@@ -223,3 +223,37 @@ def serve(destination, port, config):
...
+        descfile = os.path.splitext(target)[0] + '.md'
+    if os.path.exists(descfile) and not overwrite:
+        sys.stderr.write("Description file '{}' already exists. "
+                         "Use --overwrite to overwrite it.\n".format(descfile))
+        sys.exit(2)
+
+    with open(descfile, "w") as fp:
+        for i in range(len(keys)//2):
+            k,v = keys[i*2:(i+1)*2]
+            fp.write("{}: {}\n".format(k.capitalize(), v))
```
saimn: Maybe a print with the name of the file that has been written can be useful for users ?

saimn:
Much better, thanks :)
Could you also add a test (in `test_cli.py`) ?

tohojo:
Yup, can do; just wanted to check with you that the interface is acceptable ;)

tohojo:
And there's the test :)

saimn:
:+1:

PR #212, opened 2016-07-30
Use progressbar() instead of print() in the "Collecting albums" phase…

corsmith:
….  This removes the log spam when stdout is not a tty.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -6,6 +6,7 @@ alphabetical order):
 - Antoine Pitrou
 - Cédric Bosdonnat
 - Christophe-Marie Duquesne
+- Corey Smith (@corsmith)
 - David Siroky
 - François D. (@franek)
 - Giel van Schijndel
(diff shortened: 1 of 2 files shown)


saimn:
Hi,
Thanks for proposing this, however I think using a new progress bar for this is maybe not the best solution: we already have too many progress bars, and here we don't know by advance the number of items. For galleries with a few number of directories, it gives something that is not really useful (`Collecting albums  [#]  100%  |`).
Maybe a test could be added with `os.isatty` ?

corsmith:
I know it is a trivial change either way.  It is important to me because I run this as an automated process when new files are added to the gallery and the log spam is not great.

The progressbar in click automatically checks for isatty and the same idiom is used for each task allowing a consistent interface.

The latest commit to the feature branch removes the trailing progressChar at 100%

While collecting:
Collecting albums  [-]  0% \
When complete:
Collecting albums  [#] 100%

saimn:
Added a check with `os.isatty` in https://github.com/saimn/sigal/commit/9560f84ce465d528518690e062a239e1dc06e8cc , as I prefer not to have a new progress bar for this. Thanks anyway.

PR #213, opened 2016-08-04
Fix background-image url for default-skin.svg

datro:
PhotoSwipe icons for sharing and closing are not shown on my iPhone because the background-image url for default-skin.svg is not correct.

img_5773


saimn:
Thanks !

PR #216, opened 2016-08-20
implement a first version of video support for photoswipe theme.

franek:
CSS position is not perfect. It can be improved.

Implementation is based upon this comment : https://github.com/dimsemenov/PhotoSwipe/issues/651#issuecomment-223193536


saimn:
Thanks @franek , looks nice !
So, the only drawback is that it doesn't work with MS IE 11 and Edge 11 ?

franek:
I did not try with MS IE11 and Edge 11 but according github issue it seems. 
I try on my old Android device and it seems that it does not work too. Need to investigate.
It works well on Chrome/Firefox Desktop.

PR #224, opened 2016-12-14
Master

brentbb:
I’m using the Colorbox theme with the Media page plugin. It wasn’t properly displaying the theme on media pages because the relative path to the theme assets was wrong. This patch fixes it by changing the second argument to os.path.relpath to the album directory.


saimn:
Thanks !

PR #229, opened 2017-01-07
Allow destination directory inside the source if it is ignored

153957:
The configuration template contains the example destination `'_build'`, but that default does not work, since it is a subdirectory, which was not allowed. By allowing subdirectories (inside 'source') to be destinations, if they are ignored, the interface becomes more friendly and the default example can work.

Todo:
- [ ] Add tests
    - [ ] `exit(1)` when destination in source and not ignored
    - [ ] no error if destination in source and ignored
- [ ] Add `'_build'` as a default ignored directory

### DIFF
--- a/sigal/__init__.py
+++ b/sigal/__init__.py
@@ -23,6 +23,7 @@
 from __future__ import absolute_import, print_function
 
 import click
+import fnmatch
 import importlib
 import io
 import locale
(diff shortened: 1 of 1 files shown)


saimn:
Can you explain how you setup your gallery and why the default does not work in your case ?

153957:
Ah, I see what you mean, the default for `source` is `pictures`, I did not see that. In my case I placed the `sigal.conf.py` inside the directory containing the pictures/albums. Hence, I had it configured like so: `source = '.'` (I forgot that was my settings, and thought it was the default). Which means that with `destination = '_build'` the _build would be inside `source`.

I would still argue that this PR is still an enhancement. Allowing the build to be inside the album if the build directory is ignored should be allowed.

saimn:
>  In my case I placed the sigal.conf.py inside the directory containing the pictures/albums. Hence, I had it configured like so: source = '.'

Ok, this is typically the case I want to avoid. It should work seamlessly if you put the build directory in the ignore setting (which was added later), but still, I think it is safer to avoid this kind of setup. So I'm a bit reluctant to change this now, even if it should not be a problem.

153957:
Perhaps it could be allowed but with a warning to the user that this setup might cause issues with something like:
```python
logger.warning(
    'Your destination is inside your source directory, although it is currently '
    'ignored, this could cause problems if you ever change these settings. It is '
    'recommended to place the destination outside the source directory.')
```

saimn:
Hmm it could, but I don't see the gain of allowing it if the user needs a specific setting (i.e. make an intentional change with ignore) but then gets this message. Really I think it's better on the long term if we avoid to support this configuration (less code, less maintenance ;-)). Maybe the current error message could be more detailed ?

153957:
Ok. I can live with this, I will just need to restructure my albums a bit.
The current error message is concise and fine 😄

saimn:
Great, thanks for your feedback !

PR #231, opened 2017-01-07
Add appveyor

saimn:
Tried this out of curiosity. Currently "25 failed, 32 passed", some issues with path separators.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -14,6 +14,7 @@ platform:
     -x64
 
 install:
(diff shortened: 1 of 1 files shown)


saimn:
*"16 failed, 41 passed"*

Fixes needed for : 
- path separator
- symlink option
- locale

saimn:
Travis supports Windows now, this would be a better option if someone want to work on Windows support :)

PR #233, opened 2017-01-20
add a plugin for finer control over ignored files

t-animal:
This has the advantage that
a) you can put a .nomedia file in a directory and forget about it, no need to update the sigal config
b) you don't have to know how to write fnmatch rules
c) you can easily ignore one of two equally named directories
d) if you want to exclude many single images your config does not get cluttered
e) last (and least ;) ) you can simply copy a directory from your Android phone and .nomedia entries have the same meaning


t-animal:
Here's the docstring:

This plugin offers more fine-grained control over exluded images and folders, similarly to how
it's handled on Android.

To ignore a folder or image put a .nomedia file next to it in its parent folder and put its name
into the file.
E.g.:
```
    content of folder:
        IMG_3425.JPG, IMG_2426.JPG, IMG_2427.JPG, subfolder, .nomedia
    content of .nomedia:
        IMG_2426.JPG
        IMG_2427.JPG
        subfolder
```
will ignore all images but `IMG_3425.JPG` and the `subfolder`

Alternatively, if you put a .nomedia file into a folder and leave it blank (i.e. an empty file
called .nomedia in a folder containing images), this ignores the whole folder it's located in
(like on Android).

WARNING: When you have a pre-existing gallery from a previous run of sigal adding a new .nomedia
file will not remove the newly ignored images/albums from the existing gallery (only the entries
in the parent gallery pointing to it). They might still be reachable thereafter. Either remove
the whole gallery to be sure or remove the ignored files/folders inside the gallery to remove
them for good.

sigal/plugins/nomedia.py
```diff
@@ -0,0 +1,91 @@
...
+
+import logging
+import os
+from sigal import signals
+
+logger = logging.getLogger(__name__)
+
+def filter_nomedia(album, settings=None):
+    """Removes all filtered Media and subdirs from an Album"""
+    nomediaPath = album.src_path + "/.nomedia"
```
saimn: Use `os.path.join` instead of string concatenation.

sigal/plugins/nomedia.py
```diff
@@ -0,0 +1,91 @@
...
+
+    if os.path.isfile(nomediaPath):
+
+        if os.path.getsize(nomediaPath) == 0:
+            logger.info("Ignoring album '{}' because of present 0-byte .nomedia file".\
```
saimn: Useless backslash. Also it's better to use `logger.info('... %s ...', album.name)` as it avoids a string formatting when the logger level is higher than INFO.

sigal/plugins/nomedia.py
```diff
@@ -0,0 +1,91 @@
...
+            for key in list(album.gallery.albums.keys()):
+                if key.startswith(album.path):
+                    del album.gallery.albums[key]
+
+            #cannot set albums => empty subdirs so that no albums are generated
+            album.subdirs = []
+            album.medias = []
+
+        else:
+            with open(nomediaPath, "Ur") as nomediaFile:
```
saimn: I discover the `U` mode ;-), but from the docs: *'U' mode is deprecated and will raise an exception in future versions
of Python.  It has no effect in Python 3.* 
Also use `io.open` for Python 2 compatibility (not sure how long I will keep it !).

sigal/plugins/nomedia.py
```diff
@@ -0,0 +1,91 @@
...
+                ignoredEntries = nomediaFile.read().split("\n")
+
+                album.medias = list(filter(lambda media: media.filename not in ignoredEntries,
+                                            album.medias))
+                album.subdirs = list(filter(lambda dirName: dirName not in ignoredEntries,
+                                            album.subdirs))
+
+                #subdirs have been added to the gallery already, remove them there, too
+                for ignoredEntry in ignoredEntries:
+                    for key in list(album.gallery.albums.keys()):
```
saimn: `list` is useless here

sigal/plugins/nomedia.py
```diff
@@ -0,0 +1,91 @@
...
+                        if key.startswith(album.path + os.path.sep + ignoredEntry):
```
saimn: Use `os.path.join` instead

saimn:
> e) last (and least ;) ) you can simply copy a directory from your Android phone and .nomedia entries have the same meaning

It looks like the most convincing point to me ;-)
I know that other plugins are not tested, and that's bad 😞 , but to be sure that your plugin works it would be very useful that have some tests.

t-animal:
Ok, so I've added test cases and now the plugin also removes the ignored folders themselves from the directory tree, not just from the gallery (if they are empty) - this leaks less information.

sigal/plugins/nomedia.py
t-animal: Without the explicit list-call you'd change the dict while iterating and that throws an exception.

saimn: Good point.

saimn:
@t-animal - Sorry for the delay. It looks good, my only concern is the license of the images you added. For the other images I used personal ones or images from Wikipedia. Can you replace your images by the ones already present in the tests ? It would also save some space in the git directory (if your squash your commits).

t-animal:
The images are all from the apollo 7 mission and are in the public domain: https://www.flickr.com/photos/projectapolloarchive/21756099278/in/album-72157657129869694/

saimn:
Ok, merging ! Thanks

PR #234, opened 2017-01-20
add a progress bar for writing index files

t-animal:
This adds another progress bar. It details the progress of writing the actual `index.html`s. For large galleries (20k images in my case) this can take quite some time (30s) in my case. Without a progressbar it looks like sigal has hung up.


saimn:
Merged manually in https://github.com/saimn/sigal/commit/f4ef6ce5fabcb37ebdbc37c100f44737885ea33b. Let's see how it works with one more progress bar, thanks @t-animal.

PR #235, opened 2017-01-20
hint to how to suppress decompressionbomb warnings

t-animal:
The warnings could also be suppressed by wrapping `PILImage.open()` in Image with warnings.catch_warnings:

```
with warnings.catch_warnings():
    img = PILImage.open(...)
```

However that is not defined behaviour for multithreading: https://docs.python.org/2/library/warnings.html#temporarily-suppressing-warnings

### DIFF
--- a/sigal/image.py
+++ b/sigal/image.py
@@ -70,8 +70,11 @@ def generate_image(source, outname, settings, options=None):
         utils.copy(source, outname, symlink=settings['orig_link'])
         return
 
-    img = PILImage.open(source)
-    original_format = img.format
+    with warnings.catch_warnings(record = True) as caught_warnings:
+        img = PILImage.open(source)
+        original_format = img.format
+
+    _handle_caught_warnings(caught_warnings, source)
 
     if settings['copy_exif_data'] and settings['autorotate_images']:
         logger.warning("The 'autorotate_images' and 'copy_exif_data' settings "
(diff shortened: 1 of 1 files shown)


saimn:
Using `catch_warnings` for the `PILImage.open` calls seems more sensible, it should also catch only the `DecompressionBombWarning` (`warnings.simplefilter('ignore', Image.DecompressionBombWarning)`).
I could also be a good idea to log a message when this happens, not sure about the level, DEBUG or INFO ?

t-animal:
I updated the filter.

I don't know if it's a good idea to wrap all `PILImage.open` calls in a `catch_warnings` context, as the documentation states that "If two or more threads use the catch_warnings context manager at the same time, the behavior is undefined". I guess at least `generate_image` is called in a multithreading environment.

By default, there is a log message, because the warnings module logs to stderr.

saimn:
Sigal is using multiple processes, not threads, so it should be fine. 

> By default, there is a log message, because the warnings module logs to stderr.

But not if you catch the warning. So my idea is that it should still be possible to see the warning message in debug/verbose mode, but not in the normal mode.

t-animal:
Unfortunately one cannot use the context manager to filter for the category, so I did that manually. I logged the warnings to info, as it might be interesting to a user, not just a developer.

saimn:
> Unfortunately one cannot use the context manager to filter for the category, so I did that manually

Yep, right. Thanks !

PR #236, opened 2017-01-27
plugin to cache the exif data of images

t-animal:
Decreases the time needed to build large galleries (e.g.: 25k images in 2.5s instead of 30s)

This plugin allows extended caching, which is useful for large galleries. Once a gallery has
been built it caches the exif-data of the contained images in the gallery target folder.
Before the next run it restores them so that the image does not have to be parsed again. For
large galleries this can speed up the creation of index files dramatically.


sigal/plugins/extended_caching.py
```diff
@@ -0,0 +1,96 @@
...
+    """Restores the exif data cache from the cache file"""
+    cachePath = os.path.join(gallery.settings["destination"], ".exif_cache")
+    try:
+        if os.path.exists(cachePath):
+            with open(cachePath, "rb") as cacheFile:
+                gallery.exifCache = pickle.load(cacheFile)
+                logger.debug("Loaded cache with %d entries", len(gallery.exifCache))
+        else:
+            gallery.exifCache = {}
+    except Exception as e:
```
saimn: Maybe move this try/except around the `pickle.load` line ?

sigal/plugins/extended_caching.py
```diff
@@ -0,0 +1,96 @@
...
+        logger.warn("Could not load cache: %s", e)
+        gallery.exifCache = {}
+
+def save_cache(gallery):
+    """Stores the exif data of all images in the gallery"""
+
+    if hasattr(gallery, "exifCache"):
+        cache = gallery.exifCache
+    else:
+        cache = gallery.exifCache = {}
```
saimn: Could be simplified with ``getattr(gallery, "exifCache", {})`` and always use `cache` below.

saimn:
@t-animal - Can you rebase on master to get rid of the conflict on `tests/sample/sigal.conf.py` ?

sigal/plugins/extended_caching.py
```diff
@@ -0,0 +1,96 @@
...
+logger = logging.getLogger(__name__)
+
+def load_exif(album):
+    """Loads the exif data of all images in an album from cache"""
+    if not hasattr(album.gallery, "exifCache"):
+        _restore_cache(album.gallery)
+    cache = album.gallery.exifCache
+
+    for media in album.medias:
+        if media.type == "image":
```
saimn: Forgot this, you can use `album.images`.

saimn:
It would also be useful to add your plugins to the related documentation page, with a quick description: http://sigal.saimon.org/en/latest/plugins.html

krylovsk:
This is neat, I've been experiencing long rebuild times with my moderate (~5k) gallery.
Documentation would be welcome.

sigal/plugins/extended_caching.py
t-animal: This is intended, as the file might exist but not be readable for the user.

sigal/plugins/extended_caching.py
t-animal: mhm, I would have to change the test cases, as they assume gallery.exifCache exists...

sigal/plugins/extended_caching.py
t-animal: fixed

t-animal:
@krylovsk simply include the plugin in your config (plugins=['sigal.plugins.extended_caching']) and after you've build the gallery the next time, a cache-file will be put in your gallery from where the cached data will be loaded upon the build after that.

saimn:
Merging, thanks @t-animal !

PR #238, opened 2017-02-01
feeds: include videos in feeds

jasuarez:
Include videos when generating the feeds, besides the images.

Also, both for videos and for images without EXIF, use file modification
time.


sigal/gallery.py
```diff
@@ -186,8 +190,8 @@ class Video(Media):
     def __init__(self, filename, path, settings):
         super(Video, self).__init__(filename, path, settings)
         base, ext = splitext(filename)
-        self.date = None
         self.src_filename = filename
+        self.date = self._get_file_date() or None
```
saimn: `_get_file_date` should always return a date, so the `or None` seems useless (and the same above).

sigal/plugins/feeds.py
```diff
@@ -31,35 +31,38 @@
...
 
 def generate_feeds(gallery):
-    # Get all images and sort by date
+    # Get all images and videos and sort by date
     images = [img for album in gallery.albums.values()
               for img in album.images if img.date is not None]
-    images.sort(key=lambda im: im.date, reverse=True)
+    videos = [vid for album in gallery.albums.values()
+              for vid in album.videos if vid.date is not None]
+    medias = images + videos
```
saimn: You can use directly `album.medias`.  (`album.images` and `album.videos` are the filtered versions of the first one).

saimn:
Thanks for working on this plugin @jasuarez ! I never really finished it so it is much appreciated.

sigal/gallery.py
jasuarez: Yes, you're right. I'll remove it.

sigal/plugins/feeds.py
jasuarez: Cool!

saimn:
Thanks @jasuarez !

PR #240, opened 2017-02-20
Somes fixes/enhancements

xouillet:
This pull request contains somes fixes and enhancements:

- Add a random_thumbnail property for templates
- Enable filters in jinja templating and defined a simple datetime filter
- When using copy files, originals were always copied even if they already existed on the destination
- Same issue with zip gallery, and also add the possibility to format archive name

### DIFF
--- a/sigal/filters.py
+++ b/sigal/filters.py
@@ -1,0 +1,31 @@
+# -*- coding:utf-8 -*-
+
+# Copyright (c) 2016 - Alexandre Chataignon
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to
+# deal in the Software without restriction, including without limitation the
+# rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+# sell copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+
+from __future__ import absolute_import
+import datetime
+import locale
+
+filters = [] # Filters to be exported
+
+def date_reformat(value, format_to, format_from='%Y-%m-%d'):
+    return datetime.datetime.strptime(value, format_from).strftime(format_to).decode(locale.getpreferredencoding())
+filters.append(date_reformat)
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -29,6 +29,7 @@ import fnmatch
 import logging
 import multiprocessing
 import os
+import random
 import sys
 import zipfile
 
@@ -438,6 +439,13 @@ class Album(UnicodeMixin):
 
         self.logger.error('Thumbnail not found for %r', self)
         return None
+
+    @property
+    def rand_thumbnail(self):
+        try :
+            return join(self.name, random.choice(self.medias).thumbnail)
+        except IndexError:
+            return self.thumbnail
 
     @property
     def breadcrumb(self):
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -490,6 +490,9 @@ class Album(UnicodeMixin):
 
         if zip_gallery and len(self) > 0:
             archive_path = join(self.dst_path, zip_gallery)
+            if isfile(archive_path):
+                self.logger.debug("Archive %s already created, passing", archive_path)
+                return zip_gallery
             archive = zipfile.ZipFile(archive_path, 'w', allowZip64=True)
             attr = ('src_path' if self.settings['zip_media_format'] == 'orig'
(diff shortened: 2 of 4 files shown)


sigal/filters.py
```diff
@@ -0,0 +1,31 @@
...
+# IN THE SOFTWARE.
+
+from __future__ import absolute_import
+import datetime
+import locale
+
+filters = [] # Filters to be exported
+
+def date_reformat(value, format_to, format_from='%Y-%m-%d'):
+    return datetime.datetime.strptime(value, format_from).strftime(format_to).decode(locale.getpreferredencoding())
```
saimn: What is the goal of this filter ? There are indeed some hard-coded date formats in sigal (for exif date I think), so it may be easier to add a setting to control the date format instead of having to convert it later. 
Also the `decode` call is probably needed only for Python2 ?

sigal/writer.py
```diff
@@ -81,6 +83,9 @@ def __init__(self, settings, index_title=''):
             **env_options
         )
 
+        for filter in filters:
+            env.filters[filter.__name__] = filter
```
saimn: Having the possibility to set Jinja2 filters could be useful, but it should be possible to do it with a setting. So I would prefer that we have a list of builtin filters *not* accessible from a global var (remove `sigal.filters.filters`), and allow to add more filters with a setting.

sigal/gallery.py
```diff
@@ -101,8 +102,10 @@ def big(self):
                 return self.filename
             orig_path = join(s['destination'], self.path, s['orig_dir'])
             check_or_create_dir(orig_path)
-            copy(self.src_path, join(orig_path, self.src_filename),
-                 symlink=s['orig_link'])
+            big_path = join(orig_path, self.src_filename)
+            if not os.path.isfile(big_path):
+                copy(self.src_path, big_path,
+                     symlink=s['orig_link'])
```
saimn: Maybe add a `skip_if_exists` flag (or similar) to the `copy` function ?

sigal/gallery.py
```diff
@@ -479,7 +489,11 @@ def zip(self):
         zip_gallery = self.settings['zip_gallery']
 
         if zip_gallery and len(self) > 0:
+            zip_gallery = zip_gallery.format(album=self)
             archive_path = join(self.dst_path, zip_gallery)
+            if isfile(archive_path):
+                self.logger.debug("Archive %s already created, passing", archive_path)
+                return zip_gallery
```
saimn: This will not give the expected result if images are added to an album. Sometimes it seems reasonable to skip some operation (for the big image case, we already suppose that images are not modified, so the thumbnails are not recomputed) but here it breaks a common use case.

sigal/gallery.py
xouillet: Could be a good idea. But the main problem I had with that is that for example in the galleria template, snippets like `{%- if media.big %}<a href='{{ media.big }}'>Full size</a>{% endif %}` calls 2 times this function. Another test on `media.big` is done later so the big image is copied 4 times for nothing.

What do you think of the idea of implemeting a test with md5 hash of the file to avoid the copy and still be sure that an overwritten big image will still be correctly copied ?

sigal/gallery.py
```diff
@@ -440,6 +443,13 @@ def thumbnail(self):
         return None
 
     @property
+    def rand_thumbnail(self):
+        try :
+            return join(self.name, random.choice(self.medias).thumbnail)
+        except IndexError:
+            return self.thumbnail
```
xouillet: Ok for renaming.

This function returns thumbnail attribute so the `url_from_path` is correctly done.

sigal/gallery.py
xouillet: The problem is exactly the same than copy big image. `{% if album.zip %}...{{ album.zip }}` in the template will call the generation 2 times.

The problem here is that we don't have an easy way to detect unchanged zip without generating the zip, which is very time consuming on my collection...

sigal/filters.py
xouillet: Erm yeah, this filter in fact is useful only for me, maybe it shouldn't be in this PR.
The goal is to define some static date in my `index.md` files (global date of an album) in a standard and easy to write format and then format it via python in the template.

This is a typical case of custom filter, cf. next comment

sigal/writer.py
xouillet: To be sure we are talking about the same thing: you mean permit user to write their own filters in `sigal.conf.py` and then apply it in their template ?

As filters are heavily linked to the templates, shouldn't it be better to permit custom filters in templates directory ? For example, in a template directory, next to the `templates` and `static` folder, one can define a `filters.py` who whould contains all needed filters for this template ?

sigal/gallery.py
saimn: Right, I agree there is some nice performance improvement to gain here ;).
For the "big" image I think we only check if it exists or not, so the current patch is fine. Having a check with a md5 would be even better but it supposes that the md5 are stored somewhere.

sigal/gallery.py
saimn: Is it not meant to be used in the templates, as a url ?

sigal/gallery.py
saimn: I agree on the issue, but this case is more problematic. Without a way to know of images have been added or removed, you cannot tell if the file must be regenerated.
But there may be at least a solution to avoid generating this file multiple times. Using the `@cached_property` decorator should work.

sigal/writer.py
saimn: It could indeed be linked to the templates, but as there are not many themes/templates I think it's easier for now to put everything in `sigal.conf.py`.

sigal/gallery.py
xouillet: Yes it is, I use in my template and it works fine. I was just pointing that the call to `url_from_path` is unnecessary because this function calls `thumbnail` attribute and thus return an url path (as `url_from_path` is called in `thubnail`).

sigal/gallery.py
xouillet: Yep, `cached_property` should do the trick for now

xouillet:
As you stated, it's probably best to do multiple PR instead of this one containing multiple stuff that need reworks.

I will close this one and come back with other PRs later. Thank you for reviewing.

sigal/gallery.py
saimn: Yep but as you use a `join` it can give a bad url (mostly for Windows, on which I'm not sure if everything works correctly but ...).

saimn:
Ok, thanks !

PR #243, opened 2017-03-03
Permit filters.py file in template dir to define custom jinja filters

xouillet:


saimn:
After some thoughts on this, i think that for now it's better to allow this from the main config file, instead of having a file dedicated to filters.
Then for the future, I wonder if we should have a config file for each theme, as there are already a few settings specific to some themes. This config file could provides filters, settings, etc. that could be overridden by users in their config file. But that's probably overkill for now.

xouillet:
I've tried to put it in the config file but it is virtually impossible because of the multiprocessing that tries to serialize via pickle the configuration file.

Moreover, this solution is therefore more practical for templates developpers, far more easier to implement in the code (this PR contains less than 10 modified/added lines) and I don't see major drawbacks/regression as the current default templates doesn't use it

saimn:
> I've tried to put it in the config file but it is virtually impossible because of the multiprocessing that tries to serialize via pickle the configuration file.

Right, i forgot about this limitation 😢 
Actually because of this, and if I had the time, I would replace the Python config file by a pure text one. 

> Moreover, this solution is therefore more practical for templates developpers, far more easier to implement in the code (this PR contains less than 10 modified/added lines) and I don't see major drawbacks/regression as the current default templates doesn't use it

Yep, but that's exactly the issue, there is no mean to check that this code works just by running sigal or its test suite. That's why I was looking for a more generic solution. But ok, let's go for it as it's difficult to do better.
Can you just add some documentation, in the "Themes" page ?

sigal/writer.py
```diff
@@ -81,6 +83,13 @@ def __init__(self, settings, index_title=''):
             **env_options
         )
 
+        # handle optional filters.py
+        filters_py = os.path.join(self.theme, 'filters.py')
+        if os.path.exists(filters_py):
+            mod = imp.load_source('filters', filters_py)
+            for name in [name for name in dir(mod) if isinstance(getattr(mod, name), types.FunctionType)]:
```
saimn: The list comprehension here does not help readability:
```
for name in dir(mod):
    if isinstance(...):
```

saimn:
Merging, thanks.

PR #245, opened 2017-03-05
Improve CSP compatibility with colorbox theme

jasuarez:
This fix for colorbox theme helps to improve CSP compatibility.

Main problem is that CSP makes harder to work with static inline code, which is used in the theme.

So basically we move that code to external files, and load it in the template.

This can be considered part of fix for issue #178, at least for colorbox theme. But I think this very same idea can be applied to the other themes.


saimn:
Great, thanks @jasuarez !

PR #257, opened 2017-09-10
Set html lang attribute based upon locale

franek:
HTML lang attribute is always set to "en" for all themes. If descriptions are written in french and user agent (browser) is set in french, it may create some strange behaviors : enable browser automatic translations, speech synthesis will use english accent, ...

This PR tries to fix that. I am not a python expert so code I am publishing may surely be improved ;)

### DIFF
--- a/sigal/settings.py
+++ b/sigal/settings.py
@@ -47,6 +47,7 @@ _DEFAULT_CONFIG = {
     'index_in_url': False,
     'jpg_options': {'quality': 85, 'optimize': True, 'progressive': True},
     'keep_orig': False,
+    'html_language': 'en',
     'leaflet_provider': 'OpenStreetMap.Mapnik',
     'links': '',
     'locale': '',
--- a/sigal/templates/sigal.conf.py
+++ b/sigal/templates/sigal.conf.py
@@ -193,6 +193,9 @@ ignore_files = []
 # Specify a different locale. If set to '', the default locale is used.
 # locale = ''
 
+# Define language used on main <html> tag in templates
+# html_language = 'en'
+
 # List of files to copy from the source directory to the destination.
 # A symbolic link is used if ``orig_link`` is set to True (see above).
 # files_to_copy = (('extra/robots.txt', 'robots.txt'),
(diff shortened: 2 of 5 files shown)


sigal/gallery.py
```diff
@@ -232,6 +233,7 @@ def __init__(self, path, settings, dirnames, filenames, gallery):
         self.subdirs = dirnames
         self.output_file = settings['output_filename']
         self._thumbnail = None
+        self.language = locale.getlocale()[0][0:2]
```
franek: Really don't like that. I would appreciate a better solution.

sigal/gallery.py
```diff
@@ -232,6 +234,10 @@ def __init__(self, path, settings, dirnames, filenames, gallery):
         self.subdirs = dirnames
         self.output_file = settings['output_filename']
         self._thumbnail = None
+        
+        self.language = self.DEFAULT_LANGUAGE
+        if locale.getlocale():
+            self.language = locale.getlocale()[0][0:2]
```
franek: Really don't like that. I would appreciate a better solution.

saimn: This could simply be a setting. I would prefer this instead of relying on locale, users may want to customize it.

saimn:
Good idea to allow to customize this, but as said above it could be a setting instead.

sigal/gallery.py
franek: @saimn I thought it was a good idea to relay on locale to avoid too many settings (for instance, dates are formatting using locale). 

I will add a setting. Is ```language``` setting is enough precise or have you got a more relevant idea to name this new setting ?

saimn: Yes it's once more setting, but I think it's good to be able to choose this language, same as for the `locale` setting. 

> Is language setting is enough precise 

Not sure, it may give the feeling that the theme can be translated. I would probably go for a more specific name, maybe `html_lang` or `html_language`?

franek:
@saimn I have updated my PR. Hope it will be ok for you ;)

saimn:
Yes perfect, thanks!

franek:
\o/

PR #258, opened 2017-09-19
Resize portrait images to same size as landscape

matze:
With the current approach images are resized along the width, which means images in portrait mode are substantially smaller than landscape images. This change resizes portrait images along the height so that they will cover the exact same area.

### DIFF
--- a/sigal/image.py
+++ b/sigal/image.py
@@ -118,7 +118,13 @@ def generate_image(source, outname, settings, options=None):
             logger.error('Wrong processor name: %s', settings['img_processor'])
             sys.exit()
 
-        processor = processor_cls(*settings['img_size'], upscale=False)
+        width, height = settings['img_size']
+
+        if img.size[0] < img.size[1]:
+            # swap target size if image is in portrait mode
+            height, width = width, height
+
+        processor = processor_cls(width, height, upscale=False)
         img = processor.process(img)
 
     # signal.send() does not work here as plugins can modify the image, so we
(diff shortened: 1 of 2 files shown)


saimn:
I don't remember the exact logic behind my initial choice. It's probably that the current logic makes theming easier. With your solution you get images with different width and height.
Before implementing this in sigal, did you check if there is a similar "processor" in pilkit ? If not, I think we need a setting to change the current behavior.

matze:
> Before implementing this in sigal, did you check if there is a similar "processor" in pilkit ?

As far as I can see, pilkit does not ever checks if an image is in portrait orientation and just passes width and height verbatim.

> If not, I think we need a setting to change the current behavior.

Okay, I will add an option even though I do not see how the current behavior helps with theming.

saimn:
> Okay, I will add an option even though I do not see how the current behavior helps with theming.

It allows to have a regular grid, otherwise you will have portrait images higher than landscape ones.

matze:
The change affects the resized images not the thumbnails which still have the same aspect ratio and allow for regular grids.

saimn:
Oh indeed, sorry for the confusion ! So this choice is probably because originally there was only the galleria theme, which display images in a fixed-height frame. But I agree it would make sense to respect the orientation of the image to adjust the requested size. And after some thoughts I don't think we need a new setting for this, as there is no real drawback (only that the generated vertical images will be bigger).

So the only thing I would ask you is to update the tests to check the sizes for both horizontal and vertical images, just to be sure it works and does not break in the future.

saimn:
Thanks @matze !

PR #263, opened 2017-10-01
Add config option for tweaking how thumbnails should be cropped

3onyc:
This pull request allows specifying the `centering` parameter of [`ImageOps.fit`](pillow.readthedocs.io/en/3.1.x/reference/ImageOps.html#PIL.ImageOps.fit) in the config file, allowing for tweaks to how thumbnails are cropped with `thumb_fit = True`.

This is the most basic way I could think of implementing this, there's probably a more elegant way, and I'll gladly change this based on suggestions.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,6 +1,7 @@
 Sigal is written and maintained by Simon Conseil and contributors (in
 alphabetical order):
 
+- 3onyc
 - Abdul Qabiz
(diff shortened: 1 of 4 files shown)


saimn:
Seems good, thanks ! Could you just add yourself to the AUTHORS file ?

3onyc:
Done

saimn:
Thanks !

PR #264, opened 2017-10-02
add link to get back to toplevel page

tiberig:
Hello,
I tried sigal for the first time and I like it very much. I indexed my whole picture collection (>32k images). In order to be usable for me I needed to add a link to the top-level html page in the template I use.
I propose to merge it into the main sigal.

This is my first pull request ever. That's why I choose a small change. Pls let me know if anything more needs to be considered.
thanks,
Stefan

### DIFF
--- a/sigal/themes/photoswipe/templates/index.html
+++ b/sigal/themes/photoswipe/templates/index.html
@@ -37,6 +37,7 @@
 			{% if album.breadcrumb %}
 		    <div class="breadcrumb">
 			    <h2>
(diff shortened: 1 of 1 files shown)


saimn:
Hi,
Thanks for the proposition, but there is already a link to the root page with the page title (the site name). I think there is no need to have another link, but maybe for the photoswipe theme this link is not obvious ? I

saimn:
As discussed in #306, adding a link with a home icon would be fine. Please reopen if you want to work on this.

PR #270, opened 2017-11-06
Refactor: configurable file extensions

### LINKED ISSUE #269: Feature: configurable video extensions
It would be nice if people can add additional file extensions, so that sigal would be able to process them as videos. As an example, I'd like to use `.MTS` files that are produced by my camcorder

matelakat:
This change enables people to configure what file extensions should be
recognised as images and videos by setting:

`img_extensions` and `video_extensions`

fixes #269

### DIFF
--- a/sigal/templates/sigal.conf.py
+++ b/sigal/templates/sigal.conf.py
@@ -49,6 +49,9 @@ img_size = (800, 600)
 # show_map = False
 # leaflet_provider = 'OpenStreetMap.Mapnik'
 
+# File extensions that should be treated as images
+# img_extensions = ['.jpg', '.jpeg', '.png', '.gif']
+
 # Pilkit processor used to resize the image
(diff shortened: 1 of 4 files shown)


sigal/templates/sigal.conf.py
```diff
@@ -49,6 +49,9 @@
 # show_map = False
 # leaflet_provider = 'OpenStreetMap.Mapnik'
 
+# File extensions that should be treated as images
+# img_extensions = ['.jpg', '.jpeg', '.png', '.gif'],
```
saimn: Remove trailing comma

saimn:
Not sure why I did not go this way before, probably to limit the number of settings, but I agree that it's the easiest solution to allow users to cope with unusual extensions.

matelakat:
> Just a minor nitpick

Done @saimn

saimn:
Thanks @matelakat !

PR #273, opened 2017-11-14
Allow to configure the ffmpeg binary

leplatrem:
Can be `avconv` in certain distributions, or out of `$PATH`

### DIFF
--- a/sigal/video.py
+++ b/sigal/video.py
@@ -61,10 +61,14 @@ def check_subprocess(cmd, source, outname):
         raise SubprocessException('Failed to process ' + source)
 
 
-def video_size(source):
-    """Returns the dimensions of the video."""
+def video_size(source, settings={}):
+    """Returns the dimensions of the video.
+    :param source: path to a video
+    :param settings: settings dict
+    """
 
-    ret, stdout, stderr = call_subprocess(['ffmpeg', '-i', source])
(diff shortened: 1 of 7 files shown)


saimn:
There used to be a `ffmpeg` binary provided by avconv, is it no more the case ?

sigal/video.py
```diff
@@ -61,10 +61,14 @@ def check_subprocess(cmd, source, outname):
         raise SubprocessException('Failed to process ' + source)
 
 
-def video_size(source):
-    """Returns the dimensions of the video."""
+def video_size(source, settings={}):
```
saimn: I would use directly a `video_converter` parameter (or `converter`) here to avoid having to pass a dict just for one key.

153957: Also do not create a dict in the function definition, because if you do not provide different settings that same dict will be used each time, potentially causing problems.

From flake8-bugbear:
> B008: Do not perform calls in argument defaults. The call is performed only once at function definition time. All calls to your function will reuse the result of that definition-time call. If this is intended, assign the function call to a module-level variable and use that variable as a default value.

saimn:
Merged, thanks @leplatrem . Ideally it should be testable on Travis, but I'm too lazy right now ;)

Issue #271, opened 2017-11-13
Formatting media.datetime


nelsonov:
In the documentation for themes, you mention formatting datetime like this:
`{% if media.exif.datetime %}`
`    {{ media.exif.datetime.strftime('%A, %d. %B %Y') }}`
`{% endif %}`

However, media.exif.datetime does not appear to be a Python datetime object.  When I attempt this in the colorbox index.html, I get the following error
`File "/media/cams/themes/colorbox/templates/index.html", line 44, in template`
`    data-date=", {{ media.exif.datetime.strftime('%c') }}"`
`jinja2.exceptions.UndefinedError: 'unicode object' has no attribute 'strftime'`

Am I missing something?

Thank you.

saimn:
The documentation is wrong (sorry!), `datetime` is indeed a formatted string (with `strftime('%A, %d. %B %Y')`), but there is also a `dateobj` key with the Python datetime object:
https://github.com/saimn/sigal/blob/master/sigal/image.py#L303-L316
Can you check if it works for you ? (And if you want to fix it, you're welcome :wink: )

Issue #105, opened 2014-08-15
media.exif.datetime does not contain time


triplem:
in my locale (German), the media.exif.datetime is formatted in the following way:

Mittwoch, 23. Juli 2014

The dateobj is formatted like 2014-07-23 09:57:29, so there is a time. Basically I would expect, that the time is contained in the datetime as well.

saimn:
This one is solved, see #275.

PR #275, opened 2017-11-26
Add more media.dateobj values and correct documentation

nelsonov:
As discussed in Issue #271

In the documentation media.dateobj was mistakenly called
media.datetime while media.datetime itself is a string.

Added some more media date related values.  To avoid breaking things,
dateobj and datetime remain unchanged.  datetimelocal, datelocal and
timelocal were added.  As their names imply, they contain values
appropriate for the system's LOCALE.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -23,6 +23,7 @@ alphabetical order):
 - Julien Voisin
 - Kai Fricke
 - Keith Johnson
+- Lee Nelson
 - Lukas Vacek
 - Luke Cyca (@lukecyca)
 - Mate Lakat
--- a/docs/themes.rst
+++ b/docs/themes.rst
@@ -109,17 +109,60 @@ templates. If available, you can use:
 ``media.exif.fstop``
     The aperture value given as an F-number and formatted as a decimal.
 
-``media.exif.datetime``
+	``media.exif.dateobj``
     The time the image was *taken*. It is a datetime object, that can be
     formatted with ``strftime``:
 
     .. code-block:: jinja
 
-        {% if media.exif.datetime %}
-            {{ media.exif.datetime.strftime('%A, %d. %B %Y') }}
+        {% if media.exif.dateobj %}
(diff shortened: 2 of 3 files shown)


saimn:
Thanks for working on this! However, I think it's not necessary to introduce 3 new formatted dates (`datetimelocal` etc.) that would be computed for every images and not used most of the time. Instead, the best solution imo would be to put the format string (`'%A, %d. %B %Y'`) in a new setting. That would allow users to choose any format and also to customize the output without having to modify theme files.
BTW there is another issue about this (#105). What do you think about this solution ?

nelsonov:
#105 is the same reason I started looking into this. I'm using Sigal to post from a time-lapse weather cam, so I want to display date and time.

('%A, %d. %B %Y') is certainly understandable to us Yankees but looks odd. That's why I was leaning towards ideas like %c. 

I wouldn't want to change the behavior of media.datetime as it exists now, since other people may depend on its current functionality. I agree, however, that three extra calculations in each loop iteration is excessive.

Perhaps it's best to correct only the documentation and leave the code untouched.

saimn:
Having a new setting with the current format as default would allow to easily customize the format, and to use `%c` or whatever without having to modify theme files (which is nice for such a small customization). It's something easy to code, just add the setting with default value in `settings.py` and use it in `image.py`.

anarcat:
FWIW, `%A, %d. %B %Y` looks odd in french as well. Here it renders as `samedi, 16. décembre 2017` I'd normally write this as `samedi, 16 décembre 2017` (no comma). But I would argue that the format should be customizable or simply a [standard ISO date](https://en.wikipedia.org/wiki/ISO_8601) (where we could replace the "T" separator by a space for readability, if we display the time at all). This would look like: `2017-12-16T18:29:30-0500` or `2017-12-16 18:29:30-0500`

saimn:
I added a ``datetime_format`` setting in https://github.com/saimn/sigal/commit/1e96a7c939627d6b3750902c62f9a559cf8928d4
 Thanks anyway @nelsonov !

PR #278, opened 2017-12-12
Escape markdown body text

### LINKED ISSUE #132: Escape special characters
I have 1500 album covers scanned into iPhoto, and many of them have "quotation marks" and # hash symbols in the titles. I batch export them to the Finder using "Title" for the filenames. sigal processes them just fine. But the resulting index.html has a lot of broken thumbnails, and if I try to right-click one to inspect, it crashes Chrome 39!. At first I thought it was the sheer volume of images but no - if I rename all those images without quotes or pound symbols and re-export, it's fine. Looking at the html source, it's because the quotation marks are landing inside quoted attributes such as alt tags and file names.

So sigal should escape or convert special symbols during export.

Thanks for this great tool!

Lense:
Fix #132.

When an image description was something like

    Title: in"title"

    in"description

sigal would not escape the " in the description, generating html that
looked like

    data-title="in&#34;title"
    data-description="<br><p>in"description</p>...

This would end the attribute quote before the end of the description,
which would truncate it.

I just started using Sigal a couple hours ago, so this might break other things that I don't understand (don't think so though).

### DIFF
--- a/sigal/utils.py
+++ b/sigal/utils.py
@@ -82,7 +82,7 @@ def read_markdown(filename):
                               'markdown.extensions.tables'],
                   output_format='html5')
     # Mark HTML with Markup to prevent jinja2 autoescaping
-    output = {'description': Markup(md.convert(text))}
+    output = {'description': Markup.escape(md.convert(text))}
 
     try:
         meta = md.Meta.copy()


saimn:
You cannot escape the `description` everywhere, otherwise the HTML content would not be interpreted in the rendered page. So instead, for the case of `data-description` mentioned here, I think you have to escape only the use of `media.descrption` in the template (`{{ media.description | escape }}`).

saimn:
Closing as autoescape was activated in #195, and for galleria in https://github.com/saimn/sigal/pull/312/files#diff-a14de4e10eef7909fa269d825701b1deR137 (probably not necessary btw).

PR #280, opened 2017-12-31
include camera model and shorten exif stats

anarcat:
This adds a line to show the camera model and make, if available, and
reorders and cleans up the EXIF parameters extracted so they are more
readable for photographers (who are the ones who care about this
anyways).

Note that, as a photographer, I know what "Fstop" mean, but I would
expect "exposure" or the common f/X.Y notation. The "Focal" label
means nothing to me: "Focal length" would mean something, but it's
long - and what are the units? Just using "mm" here clarifies any
possible misunderstanding.

I have found this format to be way more readable. The following:

    ISO: 3200, Focal: 50, Exposure: 1/40, Fstop: 4.0
    Date: dimanche, 31. décembre 2017

Becomes:

    ISO 3200, 1/40s f/4.0, 50mm
    FUJIFILM X-T2
    dimanche, 31. décembre 2017

Now: dropping the explicit labels may be controversial, but I would
argue this is what markup is for. We could wrap those elements in
named <span> or <acronym> elements to provide more information through
mouse-over or CSS.

I use this patch in my install because I like to know the camera make:
it's important because I just got a new camera and I'd like to show
the difference between the different models. I'd be happy to modify to
suit your wishes better, in particular by adding the aforementioned
`<span>` elements to improve the markup.

Let me know!


saimn:
The output looks nicer indeed, but it would be good to have labels, maybe with `<abbr>` tags ?

anarcat:
okay, i'll rework this one. `<abbr>` is quite nice because it's like `<span>`; ie. it's an inline element with limited impact on layout but it can still be styled in different ways. should we have a per element "class"?

```
<abbrv class="exposure" title="exposure">f/5.3</abbr>
```

note, however, that browsers style the `<abbr>` tag differently - some add a dotted line, others in smallcaps, others with nothing at all. so we need to be careful here. see the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/abbr) for more info.

anarcat:
awesome. will do.

anarcat:
oookay... so here's the result:

```
<a href="original/DSCF0535.JPG">Full size</a>              <br>
<abbr title="film speed" class="iso">ISO 320</abbr> <abbr title="exposure" class="exposure">1/400s</abbr> <abbr title="aperture" class="aperture">f/2.8</abbr> <abbr title="focal length" class="focal">41mm</abbr>                 <br><abbr title="camera make and model" class="camera">FUJIFILM X-T2</abbr>
                <br><abbr title="date" class="date">samedi, 16. décembre 2017</abbr>
```

it's quite a mouthful, but that's the best i can lay out, because of that weird macro stuff which, btw, made it a bit hard to test; I can't use double-quotes in tags otherwise the whole page layout breaks down.

It's also hard to test this in production because the .min.css file need to be regenerated using the makefile (which, incidentally, fails if your shell is not bash because POSIX doesn't support `{foo,bar}`) and *then* reinstalled by hand. How the heck do you test this anyways?

It seems to me it would be easier if CSS would be compressed on the fly when the gallery is build instead of when the program is installed and, in any case, only if necessary... Is it so slow that we want to avoid that?

Anyways, more details about the change are in the commitlog of 48c064d9f15db49bb85970a18db865b0119decb1, but basically:

 1.  i needed to adjust the style in normalize.css... Not sure it's the right place to do that. I decided to drop the dotted line completely as it looks quite noisy, because *everything* has a dotted line in there.

 2. I also wonder if we should leave the spacing, `mm` suffixes, `f/` and `ISO` prefixes in there or we should do this in CSS

Let me know if that works for you.

saimn:
> It's also hard to test this in production because the .min.css file need to be regenerated using the makefile (which, incidentally, fails if your shell is not bash because POSIX doesn't support {foo,bar}) and then reinstalled by hand. How the heck do you test this anyways?

I do this manually but I agree it is painful :(

> It seems to me it would be easier if CSS would be compressed on the fly when the gallery is build instead of when the program is installed and, in any case, only if necessary... Is it so slow that we want to avoid that?

It's not slow, but it needs some code and an additional dependency (cssmin). I'm thinking about just removing the concatenation/minification step for now.

>    i needed to adjust the style in normalize.css... Not sure it's the right place to do that. I decided to drop the dotted line completely as it looks quite noisy, because everything has a dotted line in there.

Maybe better to put this in the theme css file to avoid losing the change when updating normalize.css.

>    I also wonder if we should leave the spacing, mm suffixes, f/ and ISO prefixes in there or we should do this in CSS

No string feelings, just keep it simple 😉

anarcat:
> It's not slow, but it needs some code and an additional dependency (cssmin). I'm thinking about just removing the concatenation/minification step for now.

i think it's still a valuable thing to do. i would do it at build time - i don't think the extra dependency is a problem: it's needed anyways...

> Maybe better to put this in the theme css file to avoid losing the change when updating normalize.css.

sorry, what's the theme css file?

> No string feelings, just keep it simple 😉

Sorry, what do you mean here? In CSS or inline?

saimn:
> i don't think the extra dependency is a problem: it's needed anyways...

Currently it (cssmin) is needed only by theme developers, not by normal users. 

> sorry, what's the theme css file?

https://github.com/saimn/sigal/blob/master/sigal/themes/galleria/static/css/style.css

> Sorry, what do you mean here? In CSS or inline?

Inline seems good enough to me, but if you think you can get a better result with some CSS, it's also fine.

anarcat:
okay, i moved it to `style.css`. i also disabled the dotted line in `abbr` by default, i found it too noisy. finally, i moved the prefix/suffix stuff (and layout!) into CSS so clients can override it as they wish. it may be a little more obscure for HTML newbies, but i think more people can figure out it and it will make this more flexible.

saimn:
Looks good, thanks !

saimn:
@anarcat - I forgot to ask you to add yourself to the AUTHOR file, I can do it but I'm not sure if you prefer to have your name or pseudonym?

anarcat:
@saimn either works, thanks!

anarcat:
i opened #291 to followup on the conversation about minifying assets.

PR #281, opened 2018-01-03
Update colorbox and photoswipe themes

saimn:


anarcat:
why was minification removed from photoswipe? it seems inconsistent with the other themes...

anarcat:
not that i mind - i propose doing exactly that for *all* themes in #291 - just curious to hear if that's where you're going with this...

saimn:
Currently I'm not going anywhere as I don't manage to find time to finish this ;), but for photoswipe I just felt that the minification step was not really required for 3 files. And I was hesitating about doing this about the other themes.

Issue #220, opened 2016-09-24
Password protected server option


Clarin98:
I am trying yo put user and pass to my Sigal installation (through apache), but I don't know how to protect the sigal port.

Have Sigal a password (user and pass) option? If it doesn't have, please add that, could be a great customization item.

Thanks

saimn:
Sigal simply generates HTML pages, there is no server process that could handle a password protection. So the password has to be handled in your web server (Apache, Nginx, ...).

Clarin98:
Do you have idea how can I generate a password in my web server (Apache)? Thanks :)

saimn:
There are many resources on this, if you search for Apache+htpassword.
Maybe https://httpd.apache.org/docs/2.2/en/howto/auth.html#gettingitworking ?

Clarin98:
Yes, thanks, but htpasswd only is useful to regular links like mydomain.com, but if you want to protect a port (like mydomain.com:8085 with username and pass) there are too few info about that on internet :(

saimn:
I don't know much about Apache configuration sorry. Closing as it is not related to Sigal.

anarcat:
So let's say I know how to configure a webserver with password protection. :) my gallery is in `/var/www/sigal/pictures` and gets built in `/var/www/sigal/_build`, as expected. I can use Apache to make `/var/www/sigal/_build` private, using a directive like:

```
    <Directory /var/www/sigal/_build>
        AuthType Basic
        AuthName "Restricted Content"
        AuthUserFile /etc/apache2/htpasswd
        Require valid-user
    </Directory>
```

So far so good: I have a private gallery - yay! But what if I want only a *part* of the gallery (say `foo/`) to be private? This is a naive solution:

```
    <Directory /var/www/sigal/_build/foo>
        AuthType Basic
        AuthName "Restricted Content"
        AuthUserFile /etc/apache2/htpasswd
        Require valid-user
    </Directory>
```

This "works" in that the images in the gallery from that directory do not get shown. However, it "fails" because the root album *also* prompts for the password, which is a terrible user experience....

So one trick I have used to workaround that is to then *whitelist* some of the files. For example, I can say that all files named `public.jpg` *can* actually be browsed even if they are in the private folder, like so:

```
    <Directory /var/www/sigal/_build/foo>
        AuthType Basic
        AuthName "Restricted Content"
        AuthUserFile /etc/apache2/htpasswd
        Require valid-user
    </Directory>

    <Files "public.jpg">
[...]

PR #285, opened 2018-01-07
introduce FAQ section

anarcat:
This is primarly made of documentation about password protection, but
will likely be expanded in the future.

See #220


saimn:
Thanks !

PR #288, opened 2018-01-17
Add compress assets plugin

Glandos:
This plugin will compress static files (html, svg, css, js) in order to be served statically by the web server.
This feature is available in nginx as `gzip_static on;`, but other web servers have similar features.

All HTTP supported encoding are supported : gzip and brotli, with gzip being able to use gzip or zopfli (gzip compatible output with a better compression).

### DIFF
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -34,7 +34,7 @@ class BaseCompressor:
         self.suffix = self.__class__.SUFFIX
 
     def compressed_filename(self, filename):
-        return '{}.{}'.format(os.path.splitext(filename)[0], self.suffix)
+        return '{}.{}'.format(filename, self.suffix)
 
     def do_compress(self, filename, compressed_filename):
         raise NotImplementedError
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -26,10 +26,10 @@ SETTINGS = {
 
 
 class BaseCompressor:
+    suffix = None
 
     def __init__(self, settings):
         self.settings = settings
-        self.suffix = self.__class__.SUFFIX
 
     def do_compress(self, filename, compressed_filename):
         '''
@@ -76,7 +76,7 @@ class BaseCompressor:
 
 
 class GZipCompressor(BaseCompressor):
-    SUFFIX = 'gz'
+    suffix = 'gz'
 
     def do_compress(self, filename, compressed_filename):
         with open(filename, 'rb') as f_in, gzip.open(compressed_filename, 'wb') as f_out:
@@ -84,7 +84,7 @@ class GZipCompressor(BaseCompressor):
 
 
 class ZopfliCompressor(BaseCompressor):
-    SUFFIX = 'gz'
+    suffix = 'gz'
 
     def do_compress(self, filename, compressed_filename):
         import zopfli.gzip
@@ -93,7 +93,7 @@ class ZopfliCompressor(BaseCompressor):
 
 
 class BrotliCompressor(BaseCompressor):
-    SUFFIX = 'br'
+    suffix = 'br'
 
     def do_compress(self, filename, compressed_filename):
         import brotli
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -136,7 +136,7 @@ def compress_assets(assets_directory, compressor):
 
 
 def compress_gallery(gallery):
-    logging.info('Compressing assets for {}'.format(gallery.title))
+    logging.info('Compressing assets for %s', gallery.title)
     settings = SETTINGS.copy()
     settings.update(gallery.settings.get('compress_assets_options', {}))
     compressor = get_compressor(settings)
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -7,8 +7,6 @@ Currently, 3 methods are supported:
 
 """
 
-from __future__ import unicode_literals
-
 import logging
 import gzip
 import shutil
@@ -62,7 +60,7 @@ class BaseCompressor:
 
         file_stats = None
         compressed_stats = None
-        compressed_filename = '{}.{}'.format(filename, self.suffix)
+        compressed_filename = u'{}.{}'.format(filename, self.suffix)
         try:
             file_stats = os.stat(filename)
             compressed_stats = os.stat(compressed_filename)
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -54,7 +54,7 @@ class BaseCompressor:
         try:
             file_stats = os.stat(filename)
             compressed_stats = os.stat(compressed_filename_result)
-        except OSError: # FileNotFoundError is for Python3 only
+        except OSError:  # FileNotFoundError is for Python3 only
             pass
 
         if file_stats and compressed_stats:
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -33,6 +33,7 @@ class BaseCompressor:
 
     def __init__(self, settings):
         self.settings = settings
+        self.suffix = self.__class__.SUFFIX
 
     def compressed_filename(self, filename):
         suffix = self.suffix()
@@ -73,9 +74,7 @@ class BaseCompressor:
 
 
 class GZipCompressor(BaseCompressor):
-
-    def suffix(self):
-        return 'gz'
+    SUFFIX = 'gz'
 
     def do_compress(self, filename, compressed_filename):
         with open(filename, 'rb') as f_in, gzip.open(compressed_filename, 'wb') as f_out:
@@ -83,9 +82,7 @@ class GZipCompressor(BaseCompressor):
 
 
 class ZopfliCompressor(BaseCompressor):
-
-    def suffix(self):
-        return 'gz'
+    SUFFIX = 'gz'
 
     def do_compress(self, filename, compressed_filename):
         with open(filename, 'rb') as f_in, open(compressed_filename, 'wb') as f_out:
@@ -93,9 +90,7 @@ class ZopfliCompressor(BaseCompressor):
 
 
 class BrotliCompressor(BaseCompressor):
-
-    def suffix(self):
-        return 'br'
+    SUFFIX = 'br'
 
     def do_compress(self, filename, compressed_filename):
         with open(filename, 'rb') as f_in, open(compressed_filename, 'wb') as f_out:
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -8,6 +8,9 @@ Currently, 3 methods are supported:
     - brotli. Need brotli module from https://pypi.python.org/pypi/Brotli. Brotli is the best compressor for web usage.
 
 """
+
+from __future__ import unicode_literals
+
 import logging
 import gzip
 import shutil
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -38,7 +38,7 @@ class BaseCompressor:
         '''
         raise NotImplementedError
 
-    def compress_file(self, filename):
+    def compress(self, filename):
         '''
         Compress a file, only if needed.
         '''
@@ -132,7 +132,7 @@ def compress_assets(assets_directory, compressor):
 
     with progressbar(assets, label='Compressing theme assets') as progress_compress:
         for filename in progress_compress:
-            compressor.compress_file(filename)
+            compressor.compress(filename)
 
 
 def compress_gallery(gallery):
@@ -146,7 +146,7 @@ def compress_gallery(gallery):
 
     with progressbar(gallery.albums.values(), label='Compressing albums static files') as progress_compress:
         for album in progress_compress:
-            compressor.compress_file(os.path.join(album.dst_path, album.output_file))
+            compressor.compress(os.path.join(album.dst_path, album.output_file))
 
     compress_assets(os.path.join(gallery.settings['destination'], 'static'), compressor)
 
--- a/sigal/plugins/compress_assets.py
+++ b/sigal/plugins/compress_assets.py
@@ -1,0 +1,157 @@
+#!/usr/bin/env python3
+
+"""Plugin to compress static files for faster HTTP transfer.
+
+Currently, 3 methods are supported:
+    - gzip. No dependency required. This is the fastest, but also largest output.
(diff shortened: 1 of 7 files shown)


saimn:
Nice idea, thanks for the PR. The main issue is that Sigal is currently still compatible with Python 2. I plan to do a final Python 2 compatible release "soon", so we can either wait before merging, or modify the code, as you like. Also even after I will want to support 3.5 so no f-strings yet, sorry!
And it would be good to have some tests.

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,157 @@
...
+    - gzip. No dependency required. This is the fastest, but also largest output.
+    - zopfli. Need zopfli module from https://pypi.python.org/pypi/zopfli. gzip compatible output with optimized size.
+    - brotli. Need brotli module from https://pypi.python.org/pypi/Brotli. Brotli is the best compressor for web usage.
+
+"""
+import logging
+import gzip
+import shutil
+import os
+import os.path
```
saimn: No need to import `os.path`.

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,157 @@
...
+        if file_stats and compressed_stats:
+            return compressed_filename_result if file_stats.st_mtime > compressed_stats.st_mtime else False
+        else:
+            return compressed_filename_result
+
+
+class GZipCompressor(BaseCompressor):
+
+    def suffix(self):
+        return 'gz'
```
saimn: `suffix` could be a class variable instead, no need for a method.

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,157 @@
...
+            f_out.write(brotli.compress(f_in.read(), mode=brotli.MODE_TEXT))
+
+
+def get_compressor(settings):
+    name = settings.get('method', '')
+    if name == 'gzip':
+        return GZipCompressor(settings)
+    elif name == 'zopfli':
+        try:
+            global zopfli
```
saimn: The use of a global is to make it available in the `do_compress` method ? That's a bit weird, you can simply import it again in `do_compress`.

Glandos: Yes, I have tried different things before doing that, but I'm not even convinced myself.
The goal is to fail fast, and to import once. I don't really like calling import on every file compression, even if it's fast.
Maybe the __init__ of each class do this job?

sigal/plugins/compress_assets.py
Glandos: Yes indeed. I'll do it with my dev environment later, it's not a thing I can do without testing ;)

Glandos:
Except for tests, is there anything more to change?

anarcat:
for what it's worth, in #291 i am proposing a similar plugin which would minify assets on build instead of in the source code directly...

Glandos:
Of course, but even minified assets can be compressed. And minification doesn't touch the generated html files also.

anarcat:
sure. i just thought you might like to know. i wonder if such a plugin should be separate or the same as the gzip plugin... ?

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
```
saimn: Can you remove this ?

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,146 @@
...
+"""Plugin to compress static files for faster HTTP transfer.
+
+Currently, 3 methods are supported:
+    - gzip. No dependency required. This is the fastest, but also largest output.
+    - zopfli. Need zopfli module from https://pypi.python.org/pypi/zopfli. gzip compatible output with optimized size.
+    - brotli. Need brotli module from https://pypi.python.org/pypi/Brotli. Brotli is the best compressor for web usage.
+
+"""
+
+from __future__ import unicode_literals
```
saimn: If you can avoid `unicode_literals`, it's often a source of hidden bugs and I prefer when proper encoding/decoding is done when needed (and I think you don't really need this here).

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,146 @@
...
+        'suffixes': ['htm', 'html', 'css', 'js', 'svg'],
+        'method': 'gzip',
+        }
+
+
+class BaseCompressor:
+
+    def __init__(self, settings):
+        self.settings = settings
+        self.suffix = self.__class__.SUFFIX
```
saimn: This is not needed. A class variable can be accessed directly via the object, so `self.suffix` works without this line. You probably just need to define it for `BaseCompressor` as well (set it to `None`).

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,146 @@
...
+        for filename in filenames:
+            assets.append(os.path.join(current_directory, filename))
+
+    with progressbar(assets, label='Compressing theme assets') as progress_compress:
+        for filename in progress_compress:
+            compressor.compress_file(filename)
+
+
+def compress_gallery(gallery):
+    logging.info('Compressing assets for {}'.format(gallery.title))
```
saimn: A minor remark, but good practice with logging is to use the "old" syntax, without explicit formatting, e.g. `logging.info('Compressing assets for %s', gallery.title)`. This way logging formats the message only if needed.

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,146 @@
...
+    settings = SETTINGS.copy()
+    settings.update(gallery.settings.get('compress_assets_options', {}))
```
saimn: I think you can avoid the copy/update here, and just use `.get` with a default value:
```
settings.get('method', SETTINGS['method'])
```
and the same for `suffixes` (and rename SETTINGS to DEFAULT_SETTINGS).

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def compressed_filename(self, filename):
+        return '{}.{}'.format(filename, self.suffix)
+
+    def do_compress(self, filename, compressed_filename):
+        raise NotImplementedError
+
+    def compress_file(self, filename):
+        compressed_filename = self.can_compress(filename)
+        if not compressed_filename:
```
saimn: I find it confusing to have the same name for a variable and a method, maybe rename the method to `get_compressed_filename` ?

saimn:
@Glandos - I left a few - mostly minor - comments. Then yes, it needs a test :wink:

sigal/plugins/compress_assets.py
Glandos: I did this to ensure a proper compatible behavior with python3. Otherwise, the migration is more painful.
However, I think I needed this because of path names that can contains non-ASCII characters. I'll have to check this fact, though.

Glandos:
Happy reviewing!
Tests will come later, then.

saimn:
@Glandos - Looks mostly good, I will have a more detailed look later. Now I just want to see that it works with some tests :wink:

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,153 @@
+"""Plugin to compress static files for faster HTTP transfer.
+
+Currently, 3 methods are supported:
+    - gzip. No dependency required. This is the fastest, but also largest output.
+    - zopfli. Need zopfli module from https://pypi.python.org/pypi/zopfli. gzip compatible output with optimized size.
+    - brotli. Need brotli module from https://pypi.python.org/pypi/Brotli. Brotli is the best compressor for web usage.
+
```
saimn: Can you wrap the lines to have something more readable ?

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,153 @@
...
+        except ImportError:
+            logging.warning('Zopfli not found, using standard gzip')
+            return GZipCompressor(settings)
+
+    elif name == 'brotli':
+        try:
+            import brotli
+            return BrotliCompressor(settings)
+        except ImportError:
+            logger.error('Unable to import brotli module')
```
saimn: Why no `GZipCompressor` in this case ?

saimn: Actually I think I prefer this, and it should do the same for zopfli. If a specific (non default) compressor is specified in the settings but the package is not found, it should always raise an error instead of using another compressor.

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,153 @@
...
+    logging.info('Compressing assets for %s', gallery.title)
+    compress_settings = gallery.settings.get('compress_assets_options', DEFAULT_SETTINGS)
+    compressor = get_compressor(compress_settings)
+
+    if compressor is None:
+        return
+
+    with progressbar(gallery.albums.values(), label='Compressing albums static files') as progress_compress:
+        for album in progress_compress:
+            compressor.compress(os.path.join(album.dst_path, album.output_file))
```
saimn: I think you could combine the two progress bars, one is enough for me. Thoughts ?

tests/test_compress_assets_plugin.py
```diff
@@ -0,0 +1,54 @@
...
+
+from sigal.gallery import Gallery
+from sigal import init_plugins
+from sigal.plugins import compress_assets
+
+
+def make_gallery(settings, tmpdir, method):
+    settings['destination'] = str(tmpdir)
+    if "sigal.plugins.compress_assets" not in settings["plugins"]:
+        settings['plugins'] += ["sigal.plugins.compress_assets"]
```
saimn: you should unregister the blinker signal at the end of the test to avoid side effects, something similar to https://github.com/saimn/sigal/commit/50072e2e2fa2f650dd1899b14aaaecb2dfe909ef#diff-56f7d7b4778bac73b6b655c02c8467aaR26

tests/test_compress_assets_plugin.py
```diff
@@ -0,0 +1,54 @@
...
+
+    suffixes = compress_options['suffixes']
+
+    for path, dirs, files in os.walk(settings['destination']):
+        for file in files:
+            path_exists = os.path.exists('{}.{}'.format(os.path.join(path, file), 'gz'))
+            file_ext = os.path.splitext(file)[1][1:]
+            assert path_exists if file_ext in suffixes else not path_exists
+
+
```
saimn: You can `@pytest.mark.parametrize` on the test above to replace the following, this dhould work:
```
@pytest.mark.parametrize("method,compress_suffix,test_import", 
                       [('gzip', 'gz', None), ('zopfli', 'gz', 'zopfli.gzip'), ('brotli', 'br', 'brotli')])
```
And you should add the zopfli & brotli packages in the .travis.yml config file, to get the tests running.

saimn:
Also, can you add some documentation in `docs/plugins.rst` and add yourself in the AUTHORS file ? Thanks!

sigal/plugins/compress_assets.py
Glandos: My thought was that zopfli is a (better) gzip compressor. So if a user wanted to use zopfli, but the module is missing, the only drawback is that files might be bigger, but everything will work as expected.
It is not the case for brotli: if the user configured its web server to serve pre-compressed brotli content, gzip files won't match.
I think that a warning is enough for zopfli, but if you insist, I can make it a real error ;)

tests/test_compress_assets_plugin.py
Glandos: Done with a fixture, inside the test itself. Feel free to copy it to conftest.py if you think it's useful.

sigal/plugins/compress_assets.py
Glandos: Yes, this is a good idea. I'll have to watch how to do it efficiently. But yes, this is clearly a better UI.

sigal/plugins/compress_assets.py
saimn: Yes I think that raising an error is better, to have a consistent behavior between options and avoid confusion for a user that will need to find why it doesn't work as expected.

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,156 @@
...
+    def get_compressed_filename(self, filename):
+        '''
+        If the given filename should be compressed, returns the compressed filename.
+        A file can be compressed if:
+            - It is a whitelisted extension
+            - The compressed file does not exist
+            - The compressed file exists by is older than the file itself
+        Otherwise, it returns False.
+        '''
+        if not os.path.splitext(filename)[1][1:] in self.suffixes_to_compress:
```
saimn: This could fail is a file has no extension ? (`splitext` returns an empty string, so slicing with `[1:]` will fail I guess)

sigal/plugins/compress_assets.py
saimn: If you merge `compress_assets` in `compress_gallery`, then:
```
with progressbar(length=len(gallery.albums) + len(assets)) as bar:
    for album in gallery.albums.values():
        ...
        bar.update()
    for filename in progress_compress:
        ...
        bar.update()
```

tests/test_compress_assets_plugin.py
saimn: Looks good, thanks !

saimn:
It seems that you picked some commits from master (they appear in the diff), did you merged master in your branch ? Could you instead rebase on master and squash your commits by the same occasion ?

sigal/plugins/compress_assets.py
```diff
@@ -0,0 +1,156 @@
+"""Plugin to compress static files for faster HTTP transfer.
+
+Currently, 3 methods are supported:
+    - gzip: No dependency required.
+            This is the fastest, but also largest output.
+    - zopfli: Need zopfli module from https://pypi.python.org/pypi/zopfli.
+              gzip compatible output with optimized size.
+    - brotli: Need brotli module from https://pypi.python.org/pypi/Brotli.
+              Brotli is the best compressor for web usage.
```
saimn: A small precision for this: if you add your plugin in `docs/plugin.rst` with autodoc (you can look other plugins in this file), then this docstring is used in the HTML doc. So it needs to be valid and well-formatted RST ;)
I can cleanup later if you prefer.

sigal/plugins/compress_assets.py
Glandos:
```
>>> os.path.splitext('test')[1][1:]
''
```
In the beginning, I thought you were right. The slicing does not failed. This is also a surprise to me 😄

Glandos:
I messed up my local branch. So delete and recreate, please go to #300. THIS. IS. COMPRESS! Sorry for that.

sigal/plugins/compress_assets.py
saimn: Oh, surprising !

PR #292, opened 2018-02-05
Colorbox add site logo in sidebar

brentbb:
This is a port of a similar feature in Pelican that allows specifying a logo filename to replace the text of the gallery name in the sidebar.


saimn:
@brentbb - Thanks, it looks useful. However I think it would be better to put the logo by default in `static/` directly (maybe not all themes have a `images/` sub-directory) and btw avoid to hard-code `images/` in the logo url. 
Also it would great if you can squash your commits to remove the merge commits, and then you can add yourself to the AUTHORS file.

saimn:
Merged manually (a102780), thanks!

PR #293, opened 2018-02-05
Colorbox use jsdelivr cdn

brentbb:
In the Colorbox theme, update jQuery to 2.2.4 and use jsdelivr for CDN to deliver third-party libraries from their latest major.minor versions


saimn:
Thanks, but I already started to update the themes (and jquery) in #281.
What would be the benefit of using jsdelivr.net ? I guess it combine the two js files ? Nowadays I'm more thinking about removing CDNs, given privacy issues, instead of adding a new one.

Glandos:
At least make CDN optional. My use case is a private gallery with a very few users, but very private data. I don't want any CDN being aware of it.
I can understand that other use cases imply a different goal.

saimn:
@Glandos - There is already a setting (`use_assets_cdn`), available for all themes (I think!), but by default CDN are used. And I'm thinking about changing this default value.

saimn:
I merged  #281 (which updates the colorbox and photoswipe themes), thanks anyway!

Issue #249, opened 2017-05-27
Use IPTC/XMP tags to display and index generation


NicoPL:
Hello,

I wonder if we could access the metadata : IPTC tags and XMP tags to use them.
IPTC (extended to XMP a while ago) are very useful for:
- Author
- Location
- Copyright
- Keywords
- etc.

1st use:
- Display some of the metadata with the picture
- Example: Display the Copyright, the Title, the Author

2nd use:
- Use them for filtering
- Example: Display according to the Author, sorted by Location

3rd use:
- when generating the gallery, be able to filter on some of the metadata.
- Example: I host all of my pictures in folders in my server. But only want to display in the gallery the ones which have the keyword 'to_display' and the copyright 'Creative Commons'

The #198 is talking a bit about the Title stored in EXIF metadata. But EXIF is not made for that normally.

Cheers

saimn:
It would be great indeed, but I'm not sure there is a Python library to read these tags ?

NicoPL:
Found the IPTCInfo library
https://pypi.python.org/pypi/IPTCInfo3/2.0.2
Seems great according to
https://twigstechtips.blogspot.co.nz/2014/06/python-reading-exif-and-iptc-tags-from.html

And the Py3Exiv2
https://launchpad.net/py3exiv2
From 2015, but maybe not as mature as the other one.

saimn:
Ok but sadly it's not so easy to find the good one: I knew pyexiv2, and the py3 version is still not pip- installable. And these two library are GPL licensed, which is not compatible with Sigal. 
It would be great to have IPTC/XMP directly in Pillow, but I don't know if there are plans for this.

NicoPL:
Right, I forgot about the licences... my mistake.
I'll look at Pillow to see if anything is planned to include IPTC tags.

saimn:
No problem, it always tricky to find good libraries (maintained, compatible license, etc.).

reagle:
This sounds related to my own request #62 .

alcoralcor:
Would this feature be implemented one day ? I need a similar feature to generate gallery with image rating.

saimn:
@alcoralcor - There is now some parsing of IPTC metadata (title, description) which are used for the images display. For other uses (e.g. filtering), it depends if someone wants to implement these kind of feature.

PR #297, opened 2018-02-13
Populate title & description from IPTC image data

edwinsteele:
The IPTC spec defines title and description fields so use those
to populate these fields on the Image unless they've been set by
some other means, thus preserving existing behaviour.

This provides a pattern that could be used with #249


sigal/gallery.py
```diff
@@ -165,6 +167,23 @@ def exif(self):
...
+            iptc_data = get_iptc_data(self.src_path)
+            if iptc_data and not self.title:
+                # 2:05 is the IPTC title property
+                if (2, 5) in iptc_data:
+                    self.title = iptc_data[(2, 5)].decode('utf-8')
+
+            if iptc_data and not self.description:
+                # 2:120 is the IPTC description property
+                if (2, 120) in iptc_data:
+                    self.description = iptc_data[(2, 120)].decode('utf-8')
```
saimn: I think you could put the details of how to get title and description in `get_iptc_data`, it would make the function easier to test and more self-contained, and avoid to do this directly in `Image` which does not have to know how IPTC work.

edwinsteele: Good idea.

saimn:
@edwinsteele - What is the license of the images ? If you are the author of these images, do you agree do share them with sigal's license ?

edwinsteele:
It’s my image (1.jpg and 2.jpg are the same image) and I’m happy to share it under an MIT licence.


--
Edwin Steele e: edwin@wordspeak.org w: https://wordspeak.org

saimn:
Great, merging then, thanks!

PR #298, opened 2018-02-14
Defer loading of leaflet js til late in the page

edwinsteele:
This is the case for other javascript. Do it for leaflet too.


saimn:
Good idea, thanks!

PR #299, opened 2018-02-15
Don't load both galleria.classic.css & minified

edwinsteele:
Fixes regression introduced in 6c5c4ed (#296)

index.html loads the galleria.classic.min.css but galleria.classic.min.js tries to load galleria.classic.css


saimn:
Are you sure that it is loaded twice ? I don't see a request for `galleria.classic.css`. I think that galleria has a way to detect that the theme is already loaded, maybe it doesn't work in your case ?

If it has to done, I would prefer not to modify galleria sources, and this is the reason why I reorganized the directory structure in #296. So we could probably use the non minified files in the template (the gain is small).

edwinsteele:
Yes, I’m sure its loaded twice (see https://www.webpagetest.org/result/180215_1A_35bcd0736678fcf9a67dbfc8ee99f59b/1/details/ where minified is loaded at 3 and non-minified at 34). That gallery is made with the current head (00f96dc).

I’ll take a look at #296 for the background. I don’t feel strongly about using the minified one in the codebase, as I can always minify as a part of my deploy process. I’m focussed on removing the duplicate request. Happy to follow your suggestion and stop using the minified one in the index.html.


--
Edwin Steele e: edwin@wordspeak.org w: https://wordspeak.org

saimn:
Oh right I missed it, sorry. Ok then, yes I prefer that we use the un-minified file in the template, to avoid modifying galleria files (otherwise the same issue may reappear the next I update galleria as I will forget about this ;) ).

saimn:
`galleria.classic.min.js` appears still modified in the final diff, although it should not. Could you maybe squash and revert this file to be sure that it is not changed ?

saimn:
Merged manually in 45ac1e4, thanks!

PR #300, opened 2018-02-15
Add compress_assets plugin

Glandos:
- The plugin, based on every conversation in #288 
- Configuration
- Myself as author
- Docs
- Tests
- Dependencies for Travis

### DIFF
--- a/tests/test_compress_assets_plugin.py
+++ b/tests/test_compress_assets_plugin.py
@@ -51,12 +51,12 @@ def make_gallery(settings, tmpdir, method):
 def test_compress(settings, tmpdir, method, compress_suffix, test_import):
     if test_import:
         pytest.importorskip(test_import)
-    compress_options = make_gallery(settings, tmpdir, 'gzip')
+    compress_options = make_gallery(settings, tmpdir, method)
 
     suffixes = compress_options['suffixes']
 
     for path, dirs, files in os.walk(settings['destination']):
         for file in files:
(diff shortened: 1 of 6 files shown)


Glandos:
I still don't understand why there is no coverage reported for brotli and zopfli. The code is tested with pytest…

saimn:
Thanks @Glandos ! From #288 I think it just misses the merging of the progress bars, not critical but it would be good to have this. I'm looking into the coverage issue.

tests/test_compress_assets_plugin.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@pytest.mark.parametrize("method,compress_suffix,test_import",
+                         [('gzip', 'gz', None),
+                          ('zopfli', 'gz', 'zopfli.gzip'),
+                          ('brotli', 'br', 'brotli')])
+def test_compress(settings, tmpdir, method, compress_suffix, test_import):
+    if test_import:
+        pytest.importorskip(test_import)
+    compress_options = make_gallery(settings, tmpdir, 'gzip')
```
saimn: So other methods are not tested because 'gzip' is always used here instead of `method` :sweat_smile:

Glandos: Ahah. You're right. This happens when it's too late to code.

Glandos:
You're right, I thought about the progress bars only after a good night. I won't edit this through GitHub though, I need to test this on my development environment.

saimn:
Thanks!

PR #302, opened 2018-02-20
[compress_assets] Mission: 100% coverage

Glandos:
This will ensure that `compress_assets` plugin is covered at 100% by tests.

To achieve that, there is a fixture that allows module masking: some module will be unavailable while the test function is running. I hope there is no side-effects.

I also set `use_orig` to True (in a832e35d) because this plugin does not care about media processing, and it really speeds tests up.

Achieving 100% coverage for this small module was a funny (but quite easy) game :)

### DIFF
--- a/tests/test_compress_assets_plugin.py
+++ b/tests/test_compress_assets_plugin.py
@@ -88,3 +88,17 @@ def test_compress(disconnect_signals, settings, tmpdir, method,
                     assert not path_exists
                 assert path_exists if file_ext in suffixes else not path_exists
 
+
+@pytest.mark.parametrize("method,compress_suffix,mask",
+                         [('zopfli', 'gz', 'zopfli.gzip'),
+                          ('brotli', 'br', 'brotli'),
+                          ('__does_not_exist__', 'br', None)])
+def test_failed_compress(mask_modules, disconnect_signals, settings, tmpdir,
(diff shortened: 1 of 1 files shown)


Glandos:
OOOOK, so python3 tests failed. Too bad.

saimn:
This looks quite complicated to gain a few percents of coverage :open_mouth: . I looked quickly and I'm not sure to sure the "meta path" thing. Could it be simpler with Python 3 features like https://docs.python.org/3/library/importlib.html#importlib.abc.MetaPathFinder ? Now that the version 1.4.0 is released, the plan is to drop Python 2, so we could maybe benefit from this.

saimn:
I wonder if we could find a simpler solution. If I understand correctly, the idea is to prevent the brotli and zopfli imports in order to test the compressor fallback. It seems there is a simpler way here: https://medium.com/python-pandemonium/how-to-test-your-imports-1461c1113be1
```
with mock.patch.dict(sys.modules, {'brotli': None}):
    ...
```

Glandos:
Yes. This is indeed a MUCH better solution, since it only involves code from the standard librairies. Mock is only available from Python 3.3, but it is OK, since 1.4 has been released.

saimn:
Thanks @Glandos !

Issue #303, opened 2018-02-21
streams page


anarcat:
Hi!

I've brought this idea up on IRC, but figured it would be more productive to discuss this formally here. :) I'd like to have a way to show the user all the recent images in the gallery. I know that the feeds plugin can somewhat do that, and even that many browsers will display that page somewhat cleanly. But it's really sub-optimal. In Firefox 58:

![image](https://user-images.githubusercontent.com/796623/36495097-320f0b7e-1702-11e8-9802-4e5c7fc93cf2.png)

First, the summary on top is all wrong: it should be HTML formatted, and instead it's one giant blob of text. Then the most important problem: the images are not clickable, which is also a bug in the feeds plugin (one which I'll get into later i guess)...

In Chromium, it looks like this:

![image](https://user-images.githubusercontent.com/796623/36495154-6315d946-1702-11e8-95c5-0f5079b82af6.png)

ie. the feed just gets downloaded. ;) There are apparently [some workarounds](https://www.labnol.org/internet/rss-feeds-in-google-chrome/24895/) for fixing that, but they mostly consist of trying to make Chrome fire up the right RSS feed reader rather than parsing the RSS feed itself (which is silly: RSS is just XML like HTML is, and should be parsable by chrome natively, but whatever).

So we could try to fight that silly battle of making web browsers respect standards, but I doubt we'll go anywhere with that. Which brings me to the idea of writing a "streams" plugin.

I started looking into that, and I've got to the point of writing a stub plugin to get my hands dirty with the mechanics of that. I blocked at the step where the Writer class expects an "Album" to write the `index.html` template:

```python
# -*- coding: utf-8 -*-

"""Plugin which adds a "stream page" listing all images.

This plugin is similar to the feeds plugin in that it allows listing
images in (reverse) chronological order, to show the most recent
ones. This allows users to browse the gallery as a "flat" list of
images, a bit like you would on sites like Flickr or Instagram.

Settigns:

- ``stream_page``, see below

Example::

    stream_page = { 'name': 'stream', 'nb_items': 10, 'title': 'Stream' }

In the above example, a 10-item page will be created at
/stream.html. If any entry is missing, a default will be used as
defined above (except `nb_items` which defaults to no limit.

.. todo:: What happens if an album with the same name exists?

"""

import logging

from sigal import signals
from sigal.writer import Writer

logger = logging.getLogger(__name__)


def generate_stream(gallery):
    logger.info('generating streams page')
    # copy-pasted from feeds.py
[...]

PR #304, opened 2018-02-21
first prototype of a stream plugin

anarcat:
The approach taken is to reuse the existing `index.html` template and
instead flatten album hierarchies in a rather unorthodox way. Media
stay linkable by prefixing the album path to their internal paths.

Many issues with the code still:

 - [ ] the way albums are flattened is definitely out of whack, with
       nasty copy() all over the place. not really sure it actually
       keeps the original albums sane. performance might break down on
       large albums.

 - [ ] class methods that should be in the Album class are just lying
       around stupidly in the plugin instead

 - [ ] code refactoring, once done, should trickle back into the feeds
       plugin as well

 - [ ] i am unsure if the modification of `filename` and
       `thumbnail_path` is sufficient for our needs. is `url`, for
       example, used anywhere?

 - [ ] no unit tests: only tested this by hand and by looking at the
       ouput.

 - [ ] documentation missing (maybe? are docstrings enough?)

Other todos:

 - [ ] only one stream is created. having per-album streams might seem
       redundant, but for large hierarchies, it could prove useful.

 - [ ] stream page is not linked from anywhere. i haven't figured out
       how this is done in the feeds plugin yet and can hack around
       this in the album descriptions for now.

 - [ ] album exclusion could prove useful to avoid the privacy leaks
       that feeds suffer from

Closes: #303 (eventually)

### DIFF
--- a/sigal/plugins/stream.py
+++ b/sigal/plugins/stream.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+
+"""Plugin which adds a "stream page" listing all images.
+
+This plugin is similar to the feeds plugin in that it allows listing
+images in (reverse) chronological order, to show the most recent
+ones. This allows users to browse the gallery as a "flat" list of
+images, a bit like you would on sites like Flickr or Instagram.
+
+Settigns:
+
+- ``stream_page``, see below
+
+Example::
+
+    stream_page = { 'filename': 'stream.html', 'nb_items': 10, 'title': 'Stream' }
+
+In the above example, a 10-item page will be created at
+/stream.html. If any entry is missing, a default will be used as
+defined above (except `nb_items` which defaults to no limit.
+
+.. todo:: What happens if an album with the same name exists?
+
+"""
+
+import copy
+import logging
+import os.path
+
+from sigal import signals
+from sigal.gallery import Album
+from sigal.writer import Writer
+
+logger = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


anarcat:
note that i have found that reordering the test suite may lead to breakage. by using `-x -ff` to repeatedly test the streams plugin, i managed to trigger a failure in `test_cli.py` (`KeyError("adjust_options")` or something). i have lost the backtrace now unfortunately, but this is something to keep in mind: maybe i introduced a regression somehow?

anarcat:
i have now mostly complete unit test coverage for the code that's actually there, and i updated the todo list in the summary. what's basically missing now is a way to link to (or include?) the stream page somewhere. the feeds plugin is embedded directly in the theme templates:

```
          {% if 'sigal.plugins.feeds' in settings.plugins %}
          {% if settings.rss_feed %}<br><a href={{ settings.rss_feed.feed_url }}>RSS Feed</a>{% endif %}
          {% if settings.atom_feed %}<br><a href={{ settings.atom_feed.feed_url }}>Atom Feed</a>{% endif %}{% endif %}</p>
```

should i just do the same, in the same place?

we also need to decide if we want a single stream page or a per-directory one, which would complicate the above template... 

i think album exclusion can be implemented in a later stage, unless you think it's absolutely necessary.

capocasa:
Just discovered this and loved it!

PR #309, opened 2018-03-06
Drop Python 2 support

### LINKED ISSUE #265: Drop Python 2 support
Once v1.4 is released (which I would like to do in a near-future), I think it will be about time to drop Python 2 support. And take advantage of Python 3 new features 😄

saimn:
Closes #265. For Sigal 2.0, Python 3.5+ will be required :tada:


requirements.txt
```diff
@@ -13,3 +14,4 @@ pytest-cov
 Sphinx
 sphinx-issues
 twine
+zopfli
```
anarcat: how does this relate to the 3.x update? are those dependencies really necessary? zopfli, for example, is not packaged in debian and will slow down inclusion a little...

.travis.yml
```diff
@@ -8,11 +8,8 @@ env:
 
 matrix:
   include:
-    - python: 2.7
-    - python: 3.4
-      env: PILLOW="Pillow==2.9.0"
     - python: 3.5
-      env: PILLOW="Pillow==3.0.0"
+      env: PILLOW="Pillow==4.0.0"
```
hugovk: How about using the latest Pillow 5.0.0?

The main reason for the major version bump was dropping Python 3.3.

https://pillow.readthedocs.io/en/latest/releasenotes/5.0.0.html

setup.py
```diff
@@ -24,14 +25,15 @@
     author=pkgmeta['__author__'],
     author_email='contact@saimon.org',
     description='Simple static gallery generator',
-    long_description=README + '\n' + CHANGELOG,
+    long_description=README,
     packages=find_packages(exclude=['tests*']),
     zip_safe=False,
     include_package_data=True,
     platforms='any',
+    python_requires='>=3.5',
```
hugovk: Good idea to include this!

It might be an idea to release a version for 2.7 and 3.4+ first (eg. 
`python_requires='>=2.7.10, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',`), see "Dropping a Python release" in https://hackernoon.com/phasing-out-python-runtimes-gracefully-956f112f33c4.

.travis.yml
anarcat: pillow 5.0 is in debian as well.

requirements.txt
saimn: Not related, I just use this file to track optional dependencies. zopfli/brotli were added recently for the compress_assets plugin (#300), but it can also work with gzip.

setup.py
saimn: Ok, thanks for the idea!

PR #310, opened 2018-03-07
Update Leaflet.js

saimn:
@anarcat - I could release a 1.4.1 version with this, if this helps Debian packaging, let me know.


anarcat:
sure, that would be useful... although it doesn't look like sigal had to be modified at all, so presumably i could have made the debian package work without the change...

i guess i didn't understood that sigal was bundling leaflet (as opposed to galleria itself)...

but it seems like a good change to do anyways, no? :)

PR #312, opened 2018-03-07
Galleria theme now reads image data from json

edwinsteele:
Having the template generate json makes it sensitive to whitespace, particularly in the description field, where failure means the gallery doesn’t load (errors in the js console). I’ve looked at ~20 galleries of mine without problems, though.

DOM interactive and document complete are noticeably improved with this change, though the first main image of the gallery is still completing its load at about the same time. To be fair the improvement will likely be more stark on a bigger gallery.

For timing information: https://webpagetest.org/video/compare.php?tests=180307_6H_80670dab108419e4b609df0c1e9a3ca3,180302_9A_ff950fed21454ed20fd5e9750d3157ca


edwinsteele:
Please hold off on merging this. I’m seeing text being written over the top of images in some cases.

edwinsteele:
The PR is ok. I suspect a recent update of galleria or leaflet started loading images from data: URLs, which were prohibited by my CSP

sigal/themes/galleria/templates/index.html
```diff
@@ -267,6 +262,20 @@ <h2>
...
+	in the background. This ensures visible elements are prioritised
+	over non-visible elements
+	
+	The choice to lazy load thumbnails in batches of 10 was not made
+	scientifically. Other numbers may yield better performance.
+	*/
+	this.lazyLoadChunks( 10 );
+    {% if settings.show_map and album.show_map %}
+    $(".gallery-map-toggle").click(galleryMapToggle);
+    {% endif %}
```
saimn: Why is this necessary or useful ? 
Also  for me the map does not show up on your branch.

sigal/themes/galleria/templates/index.html
```diff
@@ -154,6 +130,24 @@ <h2>
...
+    title: "{{ media.title if media.title else media.filename }}",
+    description: "{{ img_description(media) | e }}",
+    {% if media.type == "image" %}
+      image: "{{ media.filename }}",
+      thumb: "{{ media.thumbnail }}"
+      {%- if media.big %},
+      big: "{{ media.big }}"{% endif %}
+    {% endif %}
+    {% if media.type == "video" %}
+      video: "{{ media.filename }}"
```
saimn: So it seems that thanks to this the /img/empty.png file is not more necessary ? Do you know how galleria proceed for the thumbnail ?

sigal/themes/galleria/templates/index.html
saimn: Looking more in detail I see that it is already done a few lines above. So it should probably not been done twice ?
Actually I wonder if all the map loading could be done later, when clicking on the map button, to avoid loading tiles for nothing. But it's not related to your PR, I can try to address this later.

sigal/themes/galleria/templates/index.html
edwinsteele: I don’t have videos in my regular galleries, but when I did add one a thumbnail was generated, so I’ve made an update to use that thumbnail, and removed `/img/empty.png`

sigal/themes/galleria/templates/index.html
edwinsteele: The duplication was unintentional. Thanks for catching it. Instead of continuing to use this new `show_map` conditional, I’ve moved this simulated click into an existing conditional block that achieves the same outcome.

edwinsteele: It would be great to avoid unnecessary loading of the maps. There’s quite some data being downloaded in the background. I think the map function will be used infrequently enough that we shouldn’t unconditionally load it in the background.

edwinsteele:
> except that the map loading seems broken

I’m not exactly sure what you mean by this, but if you mean the duplication, it’s sorted now. Map loading is working for me with these changes.

sigal/themes/galleria/templates/index.html
saimn: Agreed!

saimn:
@edwinsteele - In my tests map loading was not working, I'm testing again now.

saimn:
Ok, the map works now (in your previous version it was displayed and hidden because `click(galleryMapToggle)` was called twice).

saimn:
Looks all good, thanks !

PR #316, opened 2018-03-16
Fix for crash when IPTC reading fails

IrvinDitz:
Added try/except block around iptc_data.get to catch ImageFile exceptions.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -175,11 +175,15 @@ class Image(Media):
             # Nothing to do - we already have title and description
             return
 
-        iptc_data = get_iptc_data(self.src_path)
-        if not self.title and iptc_data.get('title'):
-            self.title = iptc_data['title']
-        if not self.description and iptc_data.get('description'):
-            self.description = iptc_data['description']
+        try:
+            iptc_data = get_iptc_data(self.src_path)
+        except Exception as e:
+            self.logger.warning(u'Could not read IPTC data from %s: %s',
+                    self.src_path, e)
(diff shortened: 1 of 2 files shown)


sigal/gallery.py
```diff
@@ -175,11 +175,15 @@ def _get_metadata(self):
...
-        iptc_data = get_iptc_data(self.src_path)
-        if not self.title and iptc_data.get('title'):
-            self.title = iptc_data['title']
-        if not self.description and iptc_data.get('description'):
-            self.description = iptc_data['description']
+        try:
+            iptc_data = get_iptc_data(self.src_path)
+        except Exception as e:
+            self.logger.warning(u'Could not read IPTC data from %s: %s',
+                    self.src_path, e)
```
saimn: Could you put the following in an `else:` block, to avoid having to check if `iptc_data` is defined ?
Also the codebase is now Python 3 only, so no need to use the `u'...'` prefix for strings.

saimn:
Thanks @IrvinDitz , looks good ! Can you just address the small comments above ?

saimn:
@IrvinDitz - Can you also add yourself to the AUTHORS file ? Thanks.

sigal/gallery.py
IrvinDitz: I've left u'' for string for now as all of them logger strings are like this in the gallery. We can update all of them at once in a separate commit.

saimn: Ok, seems reasonable, thanks for the quick update!

PR #322, opened 2018-04-03
Workaround for galleria and videos

### LINKED ISSUE #321: video is not playable with galleria theme
Looks like update to newer galleria version broke video playback.

How to check:
replace theme='colorbox' => theme='galleria' in the config stored in 'tests/sample/sigal.conf.py'
run 'make demo' point browser to video testing link. The video is not playable anymore.

I did not do the proper bisect but everything was fine with git version dated by  about 2017-12-17

saimn:
Fix #321.

It looks like galleria's video tag in json support only external provider (youtube etc.). So instead this shows an empty image and on top a layer with the html video tag. 

cc @edwinsteele - do you see a better solution ?


evgmik:
Just independent confirmation. The patch fixes the non playing video issue.

saimn:
Thanks for testing @evgmik , let's go with this solution.

edwinsteele:
I don't see a better solution. Thanks for picking this up and fixing it.

PR #323, opened 2018-04-08
Inlcude tests in PyPI tarball

dotlambda:
The tests are useful for downstream packagers and do not hurt normal users (because they are usually not run).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -11,4 +11,4 @@ recursive-exclude docs *.pyo
 prune docs/_build
 recursive-include sigal/themes *
 include sigal/templates/sigal.conf.py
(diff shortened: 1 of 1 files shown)


saimn:
Sorry for the delay!
Right now I prefer not to accept this because I'm not completely sure of the license of all the images present there. It has been on my todo list to check this since a long time but ... 
Also I'm not convinced that it is useful to add ~4Mb to the package, as the majority of the users will not run the tests this way. It's not huge but if every package does the same the site-packages is quickly growing in size for no real gain.

saimn:
Ok, now that I check the included images and video (#340), let's move forward and merge this. Thanks @dotlambda !

PR #324, opened 2018-04-30
add another auto-refresh example: entr

anarcat:
Peat is great, but it's not already installed in Debian, so I am using
`entr` for the same purpose. Improve the FAQ by pointing directly at a
commandline that does the equivalent to the peat example.


saimn:
Thanks @anarcat .

PR #326, opened 2018-05-26
Continue on even when some media files are corrupt

pscripter:
I've got tens of thousands of media files in various directories. Some of them are bogus -- things like ._something.jpg, and I don't want sigal to stop just because of corrupt, or misnamed media files.

This patch fixes it.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -279,11 +279,15 @@ class Album:
 
         for f in filenames:
             ext = splitext(f)[1]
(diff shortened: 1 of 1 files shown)


saimn:
These invalid files (`._something.jpg`) should be excluded with the `ignore_files` setting instead of trying to read them. I'm reticent to adding more try/except block when there is another solution, as it make debugging more difficult.

saimn:
Closing as I could not reproduce, please reopen if this is still an issue.

PR #328, opened 2018-06-12
Remove erroneous double slash from galleria template

nebulans:
Some web servers (eg s3) will not serve correct document in this case, so remove double slash to appease.


saimn:
Thanks!

PR #343, opened 2018-10-01
Split landing page and album templates

edwinsteele:
(I suggest reviewing by commit, rather than by file - I've intentionally not squashed to help with review - I can squash if you'd like)

The templating logic in the "landing page" (the one that shows all the albums) and the "album page" has very little in common, but the two pages are currently in the same template file, which makes it more difficult to improve and debug. This PR separates the landing page and album templates. Colorbox and photoswipe has the two separate templates, but a basic include means no logic has changed, nor has any template refactoring (I'm not a user of these themes - someone could easily do the refactor, however). The two templates in Galleria have been refactored and actually have separate logic.

I have a followup branch for the galleria theme which performs prefetches of the CSS and JS resources that are used in the album pages from the landing page. This reduces the load time on the album pages. I'll open that PR if this is accepted.


sigal/writer.py
```diff
@@ -37,10 +39,8 @@
     os.path.abspath(os.path.dirname(__file__)), 'themes'))
 
 
-class Writer(object):
-    """Generate html pages for each directory of images."""
-
-    template_file = 'index.html'
+class AbstractWriter(object):
+    __metaclass__ = abc.ABCMeta
```
saimn: `__metaclass__` is a Python 2 thing, with Python 3 you have to use `class AbstractWriter(metaclass=ABCMeta):`. So it is surprising that it works. But anyway this is not needed, I would prefer the class attribute version.

sigal/plugins/media_page.py
```diff
@@ -32,15 +32,17 @@
...
 
 
-class PageWriter(Writer):
+class PageWriter(AbstractWriter):
     '''A writer for writing media pages, based on writer'''
 
-    template_file = "media.html"
+    @property
+    def template_file(self):
+        return "media.html"
```
saimn: Can you revert this change?

saimn:
Thanks for the changes, it just needs a minor change and I can merge.

Then I think the themes will need to be updated, though this can come later. One thing about the templates is that I think that we should always have a "base" template so that the different pages share the same structure. Colorbox has already this with https://github.com/saimn/sigal/blob/master/sigal/themes/colorbox/templates/base.html which is extended by the other templates.

edwinsteele:
Changes made. Good suggestion re: a base template. I'll follow the lead of Colorbox and split out the base template when I reorganise the galleria theme in the next PR.

saimn:
Ok merging, thanks!

PR #344, opened 2018-10-01
url_quote album thumbnail property

alcoralcor:
Special characters (ie. hashtags) in thumbnail url prevent them to be loaded into the page.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -409,7 +409,7 @@ class Album:
 
         if self._thumbnail:
             # stop if it is already set
-            return url_from_path(self._thumbnail)
+            return url_quote(url_from_path(self._thumbnail))
 
         # Test the thumbnail from the Markdown file.
         thumbnail = self.meta.get('thumbnail', [''])[0]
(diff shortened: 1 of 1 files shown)


saimn:
Thanks for tackling this @alcoralcor . I started something similar a long time ago (https://github.com/saimn/sigal/commit/bb5a9500f0a04b50ac5ccc14354bd8ce2e744177) and then I forgot about it :thinking: . But maybe this commit can still be useful. 
One issue since the beginning of sigal is that I did not properly separate paths and urls, and not only for thumbnails... So in https://github.com/saimn/sigal/commit/bb5a9500f0a04b50ac5ccc14354bd8ce2e744177 I started to add a `url` property, and changed the `thumbnail` property as you did, but calling `url_quote` directly in `url_from_path`, which is simpler.
So this needs more work than what you did, and the tests also needs to be updated. I can try to rebase my commit (as it was done on an old Python2-compatible version), if it is useful for you.

saimn:
Closing in favor of #345.

PR #345, opened 2018-10-04
Quote special characters in urls

saimn:
This is a quick and untested rebase of an old commit. Needs some testing.


saimn:
@alcoralcor - Do you think that you could test with this branch ? It would really useful.

alcoralcor:
I've done some tests. 
In "<img src=...", only file name is encoded. Path isn't. So, thumbnails are not loaded into page when there's a special char in path. They are loaded in last directory level because there are no more path, except "thumbnails/".
Lots of special chars have been used, all are well encoded when they are ... (in "<a href=" or file names as mentioned above).
I've tested this in all themes (galleria, colorbox and photoswipe). I'm running Python 3.6.6.

saimn:
Thanks for the testing @alcoralcor , hopefully it should be good now!

alcoralcor:
Everything's working properly now @saimn ! Thank you for the fix !

Issue #313, opened 2018-03-10
Allow easier css customization


saimn:
From #308, we could have a `user_css` setting to point to a file that would be loaded on each theme, to allow easier customization for what can be done in pure css.

PR #346, opened 2018-10-07
Squashed commits for new theme

### LINKED ISSUE #306: no "home" link in the breadcrumb
I'm currently testing Sigal/galleria

Let's say you have a folder structure like:
```
root
  AlbumA
    Sub-Album
  AlbumB
```
Once entered AlbumA, there is no way to get back to the root of the gallery, as the Breadcrumb link is limited to "AlbumA". (However it is possible to get back to AlbumA when in Sub-Album.)
I'd like to see a feature to set the "home" name of the gallery with a possibility to add it to the breadcrumb. Gallery root name could also be read from a meta file in the source folder.

Regards,

Glandos:
This theme is derived from photoswipe with the following changes:
- Use icon fonts from https://material.io/ for breadcrumb and albums.
- Remove the use of floating div, and use CSS grid instead.
- Do not put emphasis on first album media. All media are equals.
- Nicer breadcrumb with a link to home (fix #306).
- Display the caption below each media on thumbnails.
- Display the caption and the long description in enlarged view.
- Remove sharing options from enlarged view.
- In enlarged view, if the media is already zoomed at max level, a click or a tap switch to the next media.

Some questions remain:
- I have chosen to include icon fonts, but they are used for a very few icons. Is there a better option?
- This is a derived work from photoswipe theme, but it should follow some improvements from it. How can it be better achieved than manually?
- Some commits were purely personal, like the removal of sharing options.

I chose to make this a derived theme, because it displays a lot of text below media. It could have been an option of photoswipe theme, but it was way simpler to work like this. Unfortunately, I have no preview of this: I maintain a gallery of my family with it, with way too many areas to blur…

You can view some history at https://github.com/Glandos/sigal/commits/photoswipe_blog There were some rebase in it, so it's quite a mess :smile:


saimn:
Thanks for opening this @Glandos , however this raises a few questions ;).
I don't think we should have two very close themes in the package, so if that's ok for you I think we could merge some useful changes in the current photoswipe theme. Then, there should an easier way to customize theme, at least when it's doable with css only (#313, with a new setting to specify a css file), and maybe with a way to inherit a theme an override only some templates.

For icon fonts, I would prefer to avoid this in the main repo, especially if this is just for a few icons. This is because of licensing issues (for packages in distributions), need for updates, and size of the sigal package. For a few icons, using svg files instead seems simpler.

About CSS grid, is there a wide enough compatibility with browsers ?

saimn:
@Glandos - the templates have been reorganized, to make them easier to improve and maintain, see #343 and #347. Sorry for the extra work.

Glandos:
Yes, I saw them before opening this request. I know this is a test to improve my contribution quality :wink: 

Anyway, here are my thoughts on this discussion.

Derive or improve and customize
------------------------------------------
This was really an opened question for me. I chose to copy / paste so I was free to experiment a lot of thing easily, especially CSS grids and Photoswipe UI customization.
Of course, I find it far more easier to maintain only one theme instead of two, when there are only some minors differences.
But, it requires a bit more effort from me. I am quite busy, and I wanted to open the discussion, and see reactions first.

Icon font
-----------
The idea was to add some icons here and there. But I didn't know what I need then. So I put the entire font. Now that I see that I only need 2 icons, I will make them SVG.
But again, it includes some extra work. I'll do it.

CSS grid
-----------
This is the most hazardous change. All `float` and `clear` CSS directives are a bit hacky, especially with responsive design in mind, and that was exactly the goal of CSS grids. And I wanted to play with it (fun counts!).
Here is the report of [Can I Use](https://caniuse.com/#feat=css-grid). It means that it won't work with IE (I didn't use the `-ms` prefix), but it's fixable. And the only unfixable browsers are Opera Mini, Blackberry, QQ and Baidu. It seems acceptable to me, but you (and others) may not have the same expectations.

Conclusion
--------------
Please note that I removed the emphasis on the first media. It was too difficult for my CSS skills to keep it with grid (I tried, I promise), and moreover, it doesn't fit my need (make an monthly report of daily life to my far away family).

So, I will try to work as soon as possible on all these points, including merging the change from other requests.

Glandos:
Hi, I'm sorry I never finished this… But I don't want to maintain hope or something approaching it, because I won't be able to do it anyway.

I had fun doing it, but I don't have time to finish it :disappointed:

saimn:
@Glandos - No worries, I understand the lack of time to do all the nice things that could be done. Maybe another time ?

PR #349, opened 2018-10-14
GPS location code-block was incorrect, now working

subwarpspeed:
I found that the code was wrong (outdated?). It used
'media.exif.long' when the correct is 'media.exif.gps.lon'.
It was also a bit unuseful since it showed it on the most zoomed
out map (continent level) and no marker.
The current link will go to a faily zoomed in (map=18) and put
a marker in the map.


saimn:
Thanks!

Issue #284, opened 2018-01-05
watch directories


anarcat:
One thing I'm missing in my workflow is a "fire and forget" pattern. Let me explain: what I'm doing right now is I have a remote server with sigal install, where I dump pictures. Right now, I mount the `pictures` directory through SSH and dump photos there. But then I need to ssh on the remote server and run sigal by hand.

I'd like to have some sort of daemon that would sit and watch on a root directory (e.g. `~sigal/pictures` in my case) and fire up `sigal build` once changes are detected. This could be done with inotify or similar tools, but this is tricky: we don't want to start the build as soon as a file is created. What we want, actually, is to trigger the build once changes are done and no activity is occuring after a specific timeout.

I wonder if you have ever thought of implementing this: I know I could do something with `inotifywait` or similar, but I figured I would look into implementing this in sigal directly as well.

Opinions? Ideas? would a PR to that effect be welcome?

anarcat:
Fundamentally, there are two main approaches to do this in Linux land: [inotify](https://en.wikipedia.org/wiki/Inotify) and [fanotify](http://man7.org/linux/man-pages/man7/fanotify.7.html). inotify's issue is that it doesn't recursively monitor directories: you need to implement that yourself. it also takes up memory linearly with the number of monitored files. fanotify is a little better because it can recursively look into directories, but it requires root and doesn't detect renames or delete. i take this from this [excellent overview](http://www.lanedo.com/filesystem-monitoring-linux-kernel/) of the linux alternatives there.

both have stable python wrappers ([inotify](https://pypi.python.org/pypi/inotify) and [python-fanotify](https://github.com/google/python-fanotify), respectively). but they are both linux-specific, so if we worry about portability at all, we'd probably want to use a wrapper. i found those:

* [butter](https://pypi.python.org/pypi/butter) is a modern asyncio thing that does way more than just watch files, not really packaged in linux distros
* [watchdog](https://pypi.python.org/pypi/watchdog) is more minimal and completely multiplatform (including windows), [very well packaged across linux distros](https://repology.org/metapackage/python:watchdog/versions)

Then there's stuff like [gamin](https://people.gnome.org/~veillard/gamin/python.html) and [fam](https://sourceforge.net/projects/python-fam/) that have their own python bindings

[...]

PR #351, opened 2018-10-29
explain how to call sigal from git, se #284

anarcat:


saimn:
Thanks @anarcat , looks useful, but I think this deserves its own section. The current section is for a local rebuild, and with your method it is a build on a remote server. So maybe something like "How to build automatically the gallery on a remote server" ?

anarcat:
okay, makes sense, i'll rephrase

anarcat:
done

saimn:
Thanks!

PR #356, opened 2018-12-18
Extract IPTC Headline

tudacs:
As noted in the commit message, IPTC Headline (2:105) is the right field for a synopsis of the description, see also [the iptc standard](https://www.iptc.org/std/photometadata/specification/IPTC-PhotoMetadata#headline).

This pr adds the extraction of the Headline-field to `get_iptc_data()`.


saimn:
Thanks.

Issue #333, opened 2018-07-23
Symlinks in chroot


telefax:
Would it be possible to have a configuration option where you specify the symlink root?

my httpd daemon runs chrooted and the symlinks contains the full path (which the chroot doesn't know anything about).

Thanks!

telefax:
So, instead of linking to /mnt/some_mountpoint/www/images/ it would link relative to the chroot, such as /www/images/ or so.

saimn:
I understand that this behavior can be annoying, but I'm a bit reluctant to add such a specific option ... Another solution is to deactivate symlink, if your gallery is not too big the cost is small enough.

klnavajo:
I have the same problem when using the option `use_orig = True`.  Links are not relatives (but absolute path), so, images and site directories are not portable.

Perhaps a solution (`utils.py`) :

```
def copy(src, dst, symlink=False):
    """Copy or symlink the file."""
    func = os.symlink if symlink else shutil.copy2
    if symlink and os.path.lexists(dst):
        os.remove(dst)
    if symlink:  # relative symlink from dst
        func(os.path.relpath(src, os.path.dirname(dst)), dst)
    else:
        func(src, dst)
```

saimn:
> Links are not relatives (but absolute path), so, images and site directories are not portable.

Yes, but I guess that relative paths would not be so practical in many cases, as you would need to keep the same directory structure or copy the src directory with the destination one. Maybe a solution could be an option to use hardlinks, to save space on the computer building the gallery but keeping a destination directory that can be easily moved or copied ?

klnavajo:
Hi, the solution of an option and hardlink, that sounds good :)

klnavajo:
Hi, the solution of an option and hardlink, that sounds good :)

PR #359, opened 2019-01-31
Option for relative symbolic links. 

maauer:
This commit uses the code from and solves issue #333.


saimn:
Thanks! Could you add a test ? There is a `test_copy` function in `test_utils.py`.

maauer:
I did not realize that pushing would automatically add my commit. I am fixing it.

saimn:
Thanks!

PR #360, opened 2019-02-03
Add a setting to force image output format

kdmurray91:
It seems that sigal uses the format of input images for each output image. This new setting allows one to hard-code the output format of all images in the settings.conf.py.

My use case: I develop images to tiffs, and many browsers dont' like tiffs it seems. Now I can force JPEG output for all images, and browsers are happy to display them.

Thanks,
Kevin

### DIFF
--- a/sigal/templates/sigal.conf.py
+++ b/sigal/templates/sigal.conf.py
@@ -40,6 +40,9 @@ theme = 'galleria'
 # Size of resized image (default: (640, 480))
 img_size = (800, 600)
 
+# Output format of images (default: None, i.e. use input format)
+img_format = "JPEG"
+
 # Show a map of the images where possible?
 # This option only has an effect on the galleria theme for the while.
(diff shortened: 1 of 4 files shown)


kdmurray91:
(WORK IN PROGRESS)

sigal/templates/sigal.conf.py
```diff
@@ -40,6 +40,9 @@
 # Size of resized image (default: (640, 480))
 img_size = (800, 600)
 
+# Output format of images (default: None, i.e. use input format)
+img_format = "JPEG"
+
```
saimn: This should be commented as this file is meant to be used with the default values when creating a new gallery. Could you also add in the comment a list of possible values for the most common formats (e.g. png) ?

saimn:
Thanks, this can indeed be useful. If you can address my comment then I think it is good to go ?

saimn:
Thanks!

PR #362, opened 2019-02-10
Performance: Read all metadata at once

Glandos:
This will reduce the need to call PIL.Image.open, which is by far the heavier part when doing an idle build.

On my gallery with 536 images and 46 videos, I'm experiencing an improvement of ~28% faster build when writing only HTML.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -41,7 +41,7 @@ from click import get_terminal_size, progressbar
 
 from . import image, signals, video
 from .image import (get_exif_data, get_exif_tags, get_iptc_data, get_size,
-                    process_image)
+                    process_image, get_image_metadata)
 from .settings import get_thumb
 from .utils import (Devnull, cached_property, check_or_create_dir, copy,
                     get_mime, is_valid_html5_video, read_markdown,
@@ -156,6 +156,8 @@ class Media:
         self.description = ''
         """Description extracted from the Markdown <imagename>.md file."""
 
+        self.file_metadata = {}
+
         self.title = ''
         """Title extracted from the Markdown <imagename>.md file."""
 
@@ -196,6 +198,8 @@ class Image(Media):
 
     def _get_metadata(self):
         super()._get_metadata()
+        src_metadata = get_image_metadata(self.src_path)
+        self.file_metadata[self.src_path] = src_metadata
         # If a title or description hasn't been obtained by other means, look
         #  for the information in IPTC fields
         if self.title and self.description:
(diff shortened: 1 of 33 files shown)


Glandos:
The new non-covered code parts seem hard to be covered. These are `try…catch` blocks in the `get_image_metadata` after the first read. It's difficult to imagine a test case where EXIF is correct but not the size. Maybe they should be removed?

sigal/gallery.py
```diff
@@ -152,6 +152,7 @@ def _get_metadata(self):
         """ Get image metadata from filename.md: title, description, meta."""
         self.description = ''
         self.meta = {}
+        self.file_metadata = {}
```
saimn: Does it need to be a dict ? It seems that it could be an attribute of the Media object.

sigal/image.py
```diff
@@ -268,6 +272,43 @@ def get_iptc_data(filename):
...
+                       filename, e)
+
+    try:
+        iptc = get_iptc_data(img)
+    except Exception as e:
+        logger.warning('Could not read IPTC data from %s: %s',
+                       filename, e)
+
+    try:
+        size = get_size(img)
```
saimn: It seems that size is not used later ?

saimn:
Thanks @Glandos , looks like a nice improvement. About testing the exception cases, I agree that it is not easy, it would need specific images just to make the function fail. So I'm fine with not testing these cases, but we need to keep the try/catch as there are many malformed images in the wild.

sigal/image.py
Glandos: Size is not used for the source image. But it used for other images, such as thumbnails.
My goal was to code a function that extract everything that could be needed. And `get_size()` with an already opened image is really harmless from a performance point-of-view:
Performance with snakeviz
(Notice how calling `Image.open` will call `JpegImagePlugin._open` twice here).

sigal/gallery.py
Glandos: My design here is far from perfect, indeed.

I started with a simple dict, because it is flexible enough. But we could imagine a new class `MediaMetadata` that hold some functions currently in `image.py`.
But I still don't know if we need one metadata holder for each underlying file (`src`, `dst` and `thumb`), or one for all.

saimn:
Hi @Glandos ,
Where are we with this and the various changes that you (and others) made in Pillow ? I think that several key improvements were done in Pillow so is it still necessary to do the change here ? And if yes do you think that you could finalize this PR ?
Thanks

Glandos:
Thanks for the reminder. Being busy is not ideal when contributing…
Anyway, I've just run my tests with the latests Pillow (6.2.0), and this PR still improves the idle build by 12%. Not as much as before, but still significant.

So I'll try to finish this.

saimn:
No worries, I am not much available either for open source work these days.
It will be great when you can finish this one but take it easy :).

saimn:
Forgot to merge this after I had rebased...
Thanks @Glandos !

PR #363, opened 2019-02-11
Performance: Disable MPO loading

Glandos:
This one is less obvious than #362.

In [Pillow JpegImagePlugin](https://github.com/python-pillow/Pillow/blob/43ed7b29c9cffbcb05763bc29ddfc48af5db3f80/src/PIL/JpegImagePlugin.py#L778), if some headers match, then it opens the image again using `MpoImagePlugin`. In my current collection, all images are detected as such, but they are not. It leads to a double JPEG marker parsing, for nothing.

This patch disables this behavior. It can be difficult to merge for two reasons:
- It is monkey patching
- Some people might actually need to read MPO for some reason that I don't know about

In my collections of 536 images and 46 videos, I experienced a gain of 42% in speed in idle build (without applying #326). This is a huge win in this use case.

### DIFF
--- a/sigal/__init__.py
+++ b/sigal/__init__.py
@@ -34,7 +34,7 @@ from pkg_resources import get_distribution, DistributionNotFound
 from .gallery import Gallery
 from .log import init_logging
 from .settings import read_settings
-from .utils import copy
+from .utils import copy, monkey_patch_pil_jpeg_mpo
 
 try:
     __version__ = get_distribution(__name__).version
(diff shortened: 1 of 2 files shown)


saimn:
> In my current collection, all images are detected as such, but they are not.

Is this because of your images, using a wrong format, or because of the PIllow way of detecting this format ? If the latter case, it would be better to fix on their side ;).

Glandos:
See https://github.com/python-pillow/Pillow/pull/3658 for more details. It doesn't fix the `MPO` identification, but it renders its impact less visible.

Glandos:
Merged in Pillow. See https://github.com/python-pillow/Pillow/pull/3658

PR #364, opened 2019-02-11
Performance: Use the pool to collect media within albums

Glandos:
Following #362 and #363 I noticed that we could use the existing multiprocessing pool to gather media with albums.

The speed gain is not really obvious. On my gallery with 536 images and 46 videos, I'm having a 13% improvement (compared to actual master). This is not bad, but I'm also using a Samsung SSD 850 Pro 256GB, which can have good performance in multithreaded environment compared to HDD, or even some low end SSD.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -81,7 +81,6 @@ class Media:
         self.thumb_name = get_thumb(self.settings, self.filename)
         self.thumb_path = join(settings['destination'], path, self.thumb_name)
 
-        self.logger = logging.getLogger(__name__)
         self._get_metadata()
         # default: title is the filename
         if not self.title:
@@ -93,6 +92,10 @@ class Media:
 
     def __str__(self):
         return join(self.path, self.filename)
+
+    @cached_property
+    def logger(self):
+        return logging.getLogger(__name__)
 
     @property
     def url(self):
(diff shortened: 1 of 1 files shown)


sigal/gallery.py
```diff
@@ -785,3 +796,21 @@ def worker(args):
...
+    elif ext.lower() in settings['video_extensions']:
+        media = Video(f, path, settings)
+
+    # The logger is not pickable, so it cannot be used in another process.
+    # But it is a cached property, so let's delete it
+    try:
+        del media.logger
+    except AttributeError:
+        pass
+    return media
```
saimn: This should be done in the Media class, with the `__setstate__` and `__getstate__` methods: (dropping the logger in `__getstate__` and recreating it in `__setstate__`):
https://docs.python.org/3/library/pickle.html#pickle-state

sigal/gallery.py
```diff
@@ -290,17 +293,25 @@ def __init__(self, path, settings, dirnames, filenames, gallery):
...
+        if filenames:
+            if gallery.pool:
+                # Chunksize computation taken from Pool.map code
+                chunksize, extra = divmod(len(filenames), len(gallery.pool._pool) * 4)
+                if extra or chunksize < 1:
+                    chunksize += 1
+                results = gallery.pool.imap_unordered(
+                        album_worker,
+                        ((self.path, f, settings) for f in filenames),
+                        chunksize)
```
saimn: Reusing the gallery pool seems a bit fragile, not sure if this is a good idea.

saimn:
How many processes did you use to get the 13%, I guess 2 ?
Not sure if this is worth the additional complexity ...

Glandos:
I used the default, which was 4 on my system.

So yes, it is really more complex. And the gain shouldn't be so obvious on system with slow randow I/O (like HDD), or even on non-cached file.

I just present this as a potential improvement for some use cases.

sigal/gallery.py
Glandos: It's not as fragile as the code is _right now_. All albums are collected before the actual processing, so the pool should be free when we need it again.

saimn:
Same questions as the other PR, is this impacted by the Pillow changes ?

Glandos:
Improvements in Pillow reduced the impact of these changes. And since there are very intrusive, let's close this, and keep it for later reference for the time when everyone has at least 12-core CPU ;)

saimn:
Thanks for the update!

PR #367, opened 2019-03-09
360 photo and video support

mittman:
Basic integration of Valiant360 webGL library to view 360-degree photos and videos within sigal gallery. Modification of galleria theme, click VR icon or on the active image to open modal.


saimn:
Hi,
Thanks for the PR, this seems to be a nice feature. However I'm not sure to want this in the core galleria theme, because this means more dependencies, and for each release it is a lot of work to update all the dependencies to their latest versions and check that everything still works.
What would think about putting this in another repository, like the [153957-theme](https://github.com/153957/153957-theme/) (which can be pip-installed, which is great but not strictly necessary) ? The theme could then be referenced in https://github.com/saimn/sigal/wiki with other community efforts.

PR #368, opened 2019-05-14
Add new plugin to allow one to disable ZIP generation on a per album basis

riton:
Heavily inspired by the `.nomedia` file, this patch introduces a new plugin called `nozip_gallery`.

If `zip_gallery` is _enabled_, this plugin allows to control on a per album basis if a ZIP file should be generated or not.

It currently uses a `.nozip_gallery` file (empty or not) to achieve such behavior.

The motivation behind this plugin is:
* disable ZIP file generation for my _very old_ photo albums
* enable ZIP file generation for my recent photo albums

### DIFF
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -119,6 +119,11 @@ Nomedia plugin
 
 .. automodule:: sigal.plugins.nomedia
 
+No ZIP Gallery plugin
+=====================
+
+.. automodule:: sigal.plugins.nozip_gallery
+
 Upload to S3 plugin
 ===================
 
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -2,6 +2,8 @@
  Plugins
 =========
 
+.. contents::
+
 How to use plugins
 ------------------
 
@@ -119,11 +121,6 @@ Nomedia plugin
 
 .. automodule:: sigal.plugins.nomedia
 
-ZIP Gallery plugin
-==================
-
-.. automodule:: sigal.plugins.zip_gallery
-
 Upload to S3 plugin
 ===================
 
@@ -133,3 +130,8 @@ Watermark plugin
 ================
 
 .. automodule:: sigal.plugins.watermark
+
+ZIP Gallery plugin
+==================
+
+.. automodule:: sigal.plugins.zip_gallery
--- a/docs/plugins.rst
+++ b/docs/plugins.rst
@@ -119,10 +119,10 @@ Nomedia plugin
 
 .. automodule:: sigal.plugins.nomedia
 
-No ZIP Gallery plugin
-=====================
+ZIP Gallery plugin
+==================
 
-.. automodule:: sigal.plugins.nozip_gallery
+.. automodule:: sigal.plugins.zip_gallery
 
 Upload to S3 plugin
 ===================
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -267,6 +267,7 @@ class Album:
         self.settings = settings
         self.subdirs = dirnames
         self.output_file = settings['output_filename']
+        self.disable_zip_gallery = False
         self._thumbnail = None
 
         if path == '.':
@@ -527,6 +528,12 @@ class Album:
         archive with all original images of the corresponding directory.
 
         """
+
+        if self.disable_zip_gallery:
+            # ZIP file generation has been explicitly disabled
+            # by a .nozip_gallery file for instance
+            return
+
         zip_gallery = self.settings['zip_gallery']
 
         if zip_gallery and len(self) > 0:
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -31,7 +31,6 @@ import os
 import pickle
 import random
 import sys
-import zipfile
 
 from click import progressbar, get_terminal_size
 from collections import defaultdict
@@ -267,7 +266,6 @@ class Album:
         self.settings = settings
         self.subdirs = dirnames
         self.output_file = settings['output_filename']
-        self.disable_zip_gallery = False
         self._thumbnail = None
 
         if path == '.':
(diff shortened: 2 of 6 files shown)


riton:
__Note__: 

Another solution would have been to rely on the album `index.md` file with a `GenerateZIP: true / false` setting.

I've chosen to use a `.nozip_gallery` file like the `.nomedia` file for consistency.

saimn:
Hi,
Thanks for the PR, and sorry for the delay. This would indeed be a nice feature to have. Thinking about it, we could maybe move the zip feature to the plugin, that would then take care of adding the zip method to the albums when the plugin is activated and when the special file does not exists. Would you be interested to have a look at this?
Also it will need some unit tests :)

riton:
Hi, my turn to take a long time to respond, sorry about that.

I've made some modifications and moved the whole _zip generation_ logic to a `zip_gallery` plugin.
Is this what you meant / wanted ?

About the unit tests, I'll see if I can find some time to write them. Before going further, I want to be sure that we're on the same page about the _zip logic_ modifications that I've made in fc14de3872a0fdf26e1c366fa7bdf287ea1ad69d

riton:
BTW @saimn , just noticed that you're from Lyon. If you want to meet and discuss further in real life next to a beer or so, I'd be glad :wink:

sigal/plugins/zip_gallery.py
```diff
@@ -0,0 +1,90 @@
...
+    nozipgallerypath = os.path.join(album.src_path, ".nozip_gallery")
+
+    zip_method = generate_album_zip_method
+
+    if os.path.isfile(nozipgallerypath):
+        logger.info("Ignoring ZIP gallery generation for album '%s' because of present "
+                    ".nozip_gallery file", album.name)
+        zip_method = noop_zip_method
+
+    album.__class__.zip = cached_property(zip_method)
```
saimn: Since the property is set on the class, it will be used for all the Album objects (the last will win). So instead I think that you can check for the "nozip" file directly in `generate_album_zip_method` ?

sigal/gallery.py
```diff
@@ -521,37 +520,10 @@ def show_map(self):
...
-                path = getattr(p, attr)
-                try:
-                    archive.write(path, os.path.split(path)[1])
-                except OSError as e:
-                    self.logger.warn('Failed to add %s to the ZIP: %s', p, e)
-
-            archive.close()
-            self.logger.debug('Created ZIP archive %s', archive_path)
-            return zip_gallery
+        return None
```
saimn: I wonder if having the placeholder method is necessary? Jinja evaluates missing attributes to an empty string - I think, so it should be fine if the `.zip` method is missing.

sigal/plugins/zip_gallery.py
```diff
@@ -0,0 +1,90 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+
+""" This plugin controls the generation of a ZIP archive for a gallery
+
+To ignore a ZIP gallery generation for a particular album, put a ``.nozip_gallery`` file next to it in its parent
+folder. Only the existence of this ``.nozip_gallery`` file is tested.
```
saimn: Could you mention here the `zip_gallery` setting, as this will be shown in the docs. Basically what's in the former `zip` docstring.

saimn:
Thanks @riton , I think this is a very nice improvement! 
For the test, there is already a `test_zip.py` file, so the easiest would to add a "nozip" file in some directory in `tests/sample/pictures`, and check that the zip archive for this directory is not created (in `test_zipped_correctly`).
And yes, I would be happy to have a beer in Lyon :)

sigal/plugins/zip_gallery.py
riton: You're absolutely right, this was too naive. I've switched to a single _function_ that does all the checks and contains all the logic.

sigal/gallery.py
riton: I don't think this is _necessary_. I just thought this can make the code _easier to read_.
For the reader, this _placeholder_ may give a hint on where the actual implementation of the _zip_ property is made. (I'm not a huge fan of _magic properties that just appears you don't know where from_ even if python allows it :smile: )

Anyway, If you'd like me to remove the placeholder, no problem for me. I can simply replace it with a comment to help future readers.

sigal/plugins/zip_gallery.py
riton: I've adapted the _docstring_. Should be okay now.

riton:
@saimn  I've made some quick modifications following your suggestions. Still no unit tests for now.

docs/plugins.rst
```diff
@@ -119,6 +119,11 @@ Nomedia plugin
 
 .. automodule:: sigal.plugins.nomedia
 
+No ZIP Gallery plugin
```
saimn:
```suggestion
ZIP Gallery plugin
```
The plugin is more general now ;)

sigal/gallery.py
saimn: I see your point about the magic aspect and agree in general, but on the other side it is weird to have some unrelated zip stuff here, since this is now an independent feature. And the implementation will be easy to find in the zip plugin.

sigal/plugins/zip_gallery.py
```diff
@@ -0,0 +1,103 @@
...
+        return False
+
+    return _generate_album_zip(album)
+
+def nozip_galery_file(album, settings=None):
+    """Filesystem based switch to disable ZIP generation for an Album"""
+    Album.zip = cached_property(generate_album_zip)
+
+def register(settings):
+    signals.album_initialized.connect(nozip_galery_file)
```
saimn: galery -> gallery (and the same above)

sigal/plugins/zip_gallery.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for p in album:
+            path = getattr(p, attr)
+            try:
+                archive.write(path, os.path.split(path)[1])
+            except OSError as e:
+                album.logger.warn('Failed to add %s to the ZIP: %s', p, e)
+
+        archive.close()
+        album.logger.debug('Created ZIP archive %s', archive_path)
```
saimn: You can use the `logger` defined at the top of the module instead of `album.logger`.

saimn:
@riton - Could you have a look at the test to see why it is failing ?

saimn:
Hi @riton ,
I have finalized the PR with the tests and a few details, so this is now good to go. Thanks for your work !

PR #374, opened 2019-09-01
getting started doc: add "album." qualifier where needed

zacchiro:
without this, "meta" is not accessible


saimn:
Thanks @zacchiro !

PR #377, opened 2019-10-17
Filter by metadata

stasinos:
Adds the feature to use XMP tags to select the images that will be included in the gallery.


Glandos:
If I remember correctly, `pyexiv2` was deliberately not used, in the name of ease of installation.
I might be wrong, @saimn will confirm this or not.

I know that it is a powerful library, and that metadata is not well supported with Pillow.

stasinos:
I didn't have to do anything more than pip install, but maybe I was just lucky with glibc versions and such, travis here doesn't look happy. If that is a deciding factor, I can try to make this conditional so that pyexiv2-less installations are just as easy as they are now, but ignore the tag filters.

saimn:
Until recently my knowledge of it was that it was not pip installable, and there was a [fork](https://python3-exiv2.readthedocs.io/en/latest/) for python 3 support. But maybe things have changed, which can be great. I will try to look more in detail soon.

saimn:
And even if the installation is easier, I think that this feature would fit well in a plugin.

saimn:
Hmm it's a bit confusing:
- the pyexiv2 I know about is https://launchpad.net/pyexiv2, it is packaged on Debian (https://packages.debian.org/fr/sid/python-pyexiv2), and it is deprecated in favour of GExiv2 (https://wiki.gnome.org/Projects/gexiv2)...
- PyPI's pyexiv2 is a different project, a new one, with a first release in July 2019. https://github.com/LeoHsiao1/pyexiv2

stasinos:
I wasn't aware of GExiv2, and it looks more likely to offer a hussle-free installation path. The latest commit switches to using that instead.

Not sure about how your plugins work, though. Any pointers?

saimn:
Gexiv2 may be easier to install on Linux, but probably not on MacOS/Windows. There is no package on PyPI.

If I look at PyPI (https://pypi.org/search/?q=exiv2&o=), there is [py3exiv2](https://pypi.org/project/py3exiv2/) which is the py3 fork on the old pyexiv2, but it does not provide wheels so you need to install libexiv2-dev and libboost-python-dev and compile. And there is the ["new pyexiv2"](https://pypi.org/project/pyexiv2/), which provides a universal wheel which is weird for a package containing C code, and it seems that .dll and .so are committed directly in the repository ..

So it's a mess and I do not want this in the core package ;), but if you manage to write a plugin that can work with one or some of these packages then it's fine :)

About plugins: http://sigal.saimon.org/en/latest/plugins.html
You may need a new entry point (signal).

PR #384, opened 2020-02-29
Improve error message when template is not found

t-animal:
The previous error message did not include the path that was
searched in. By including the path the debugging of the config
becomes way easiert (e.g. in my case I had added the '/templates'
folder which was added again by sigal).


saimn:
Thanks :+1:

PR #385, opened 2020-03-30
Plugin encrypt

dbw9580:
This plugin adds support for password-protected galleries. Images are encrypted using a password during the build, and then decrypted in browser when they are accessed by viewers if they can provide the correct password.

Summaries:

* Pure Javascript implementaion thanks to Web Crypto, no server application or database is needed;
* Images are encrypted with AES-128-GCM, keys are generated using PBKDF2 with SHA1;
* Support incremental builds;
* `decrypt.js` is tested to work with all 3 themes bundled with Sigal, on Chrome and Firefox;


dbw9580:
Demo: https://dbw9580.github.io/sigal-encrypt-demo/

saimn:
Looks cool, thanks!
The code seems good, and  Ihave not much to say about the encryption part, maybe just a few questions.
I wonder if we could simplify a bit your implementation by adding some signals in sigal's code, for instance when writing an image, which could be used to write the encrypted version without having to check if it was already encrypted or not. It would also be useful to have a way to inject the HTML code in the header when writing HTML files, I think templates have an extra_head tag (or similar) so if there was a signal before writing HTML fies which you could use to inject your code it would also simplify things I guess.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,297 @@
...
+
+def cache_key(media):
+    return os.path.join(media.path, media.filename)
+
+def save_property(cache, media):
+    key = cache_key(media)
+    if key not in cache:
+        cache[key] = {}
+        cache[key]["size"] = media.size
+        cache[key]["thumb_size"] = media.thumb_size
```
saimn, line 130: What are those useful for ? (size and thumb_size)

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,297 @@
...
+                head.append(scriptNode)
+                head.append(scriptConfig)
+                with open(index_file, "w", encoding="utf-8") as f:
+                    f.write(root.prettify())
+
+def copy_assets(settings):
+    theme_path = os.path.join(settings["destination"], 'static')
+    for root, dirs, files in os.walk(ASSETS_PATH):
+        for file in files:
+            copy(os.path.join(root, file), theme_path, symlink=False, rellink=False)
```
saimn, line 291: You have only two files to copy, so maybe use their path directly here? Would be more explicit.

tests/sample/sigal.conf.py
```diff
@@ -17,7 +17,16 @@
...
+    'sigal.plugins.encrypt'
 ]
+encrypt_options = {
+    'password': 'password',
+    'ask_password': True,
+    'gcm_tag': 'AuTheNTiCatIoNtAG',
+    'kdf_salt': 'saltysaltsweetysweet',
+    'kdf_iters': 10000,
+    'encrypt_symlinked_originals': False
+}
```
saimn, line 29: Maybe not add this by default, since this config is used by several tests.

saimn:
Do you think you could add some tests, building the gallery with the example settings and checking that images are indeed encrypted?

dbw9580:
> simplify a bit your implementation by adding some signals in sigal's code

Definitely a good idea. I'll work on that.

> add some tests, building the gallery with the example settings and checking that images are indeed encrypted

So I was trying to create some tests, but not sure about how fine granularity it is suitable for such a plugin. So basically what the test should look like is first build the sample gallery, encrypt several pictures and decrypt them in-place to see if decryption works out with no error, right? As for the javascript parts of code, I am not sure how to test them properly.

sigal/plugins/encrypt/encrypt.py
dbw9580: They are cached for future incremental builds because once images are encrypted, they are not readable by PIL to get the sizes. Just like how the `extended_caching` plugin works.

dbw9580: Also I think the `extended_caching` plugin should be enabled when using `encrypt`, for the same reason. I will update the documentation to reflect that.

sigal/plugins/encrypt/encrypt.py
dbw9580: Sure.

tests/sample/sigal.conf.py
dbw9580: I'm not quite familiar with the testing infrastructure. So the plugin options should be hardcoded in the test files? Should this plugin be enabled in the test config by default?

sigal/plugins/encrypt/encrypt.py
dbw9580: Changed in 181d0bb

setup.cfg
```diff
@@ -37,7 +37,7 @@ install_requires =
     pilkit
 
 [options.extras_require]
-all = boto; brotli; feedgenerator; zopfli
+all = boto; brotli; feedgenerator; zopfli; cryptography; beautifulsoup4
```
dbw9580, line 40: Since JS files are now injected via Jinja templates, `beautifulsoup4` is no longer a dependency.
Note to self: remove bs4 from code.

saimn:
> So I was trying to create some tests, but not sure about how fine granularity it is suitable for such a plugin. So basically what the test should look like is first build the sample gallery, encrypt several pictures and decrypt them in-place to see if decryption works out with no error, right?

Yes, something like this. Testing that some images are indeed encrypted, and maybe that they can be descrypted with the Python decrypt code that you included ?
There is an example of how to build the test gallery and modify the settings here:
https://github.com/saimn/sigal/blob/0d8a3c8779ab32855575bbdafba02645d4b0e290/tests/test_plugins.py#L9-L19

> As for the javascript parts of code, I am not sure how to test them properly.

Yeah that's doable but it's a lot of work so I never went into this.

dbw9580:
I rewrote most of the frontend code with Service Worker. The decryption process is now (almost) transparent to the gallery app and the user. Image `src`s are no longer replaced with `blob:` urls, which means many browser extensions that rely on these links to work (e.g. image bulk downloaders, thumbnail previewers) can now function properly. Also original images can be opened in a separate tab without showing an error (which was impossible with the previous approach that based on hijacking image load events). 

Speaking of original images, I find that they are handled differently than the resized ones and the thumbnails, in terms of how they end up where they belong in the output directory.

They are copied here: 
https://github.com/saimn/sigal/blob/0d8a3c8779ab32855575bbdafba02645d4b0e290/sigal/gallery.py#L120

while the resized and thumbnails are directly saved from PIL here:
https://github.com/saimn/sigal/blob/0d8a3c8779ab32855575bbdafba02645d4b0e290/sigal/image.py#L144

and here:
https://github.com/saimn/sigal/blob/0d8a3c8779ab32855575bbdafba02645d4b0e290/sigal/image.py#L163

This means if I were to use signals to handle the encryption, I would need one signal for the original files, and another for the resized and thumbnails. Which seems a little overkill to me.

Your thoughts?

setup.cfg
dbw9580: Removed in 28f187f.

tests/sample/sigal.conf.py
dbw9580: Reverted in 28f187f.

dbw9580:
@saimn I've added the test. I think this PR is ready for review.

sigal/writer.py
```diff
@@ -112,8 +113,10 @@ def generate_context(self, album):
 
     def write(self, album):
         """Generate the HTML page and save it."""
-
-        page = self.template.render(**self.generate_context(album))
+        context = self.generate_context(album)
+        for receiver in signals.before_render.receivers_for(context):
+            context = receiver(context)
```
saimn, line 118: Any reason why you need to iterate on receivers instead on sending context directly ?
```suggestion
        signals.before_render.send(context)
```

tests/test_encrypt.py
```diff
@@ -0,0 +1,91 @@
...
+from sigal.plugins.encrypt.encrypt import cache_key
+
+CURRENT_DIR = os.path.dirname(__file__)
+
+
+@pytest.fixture()
+def remove_cache(settings):
+    yield
+    cachepath = os.path.join(settings['destination'], ".encryptCache")
+    os.remove(cachepath)
```
saimn, line 18: As the cache goes in a temp dir, why do you need to remove it ?

sigal/writer.py
dbw9580: Right... I was looking at here

https://github.com/saimn/sigal/blob/0d8a3c8779ab32855575bbdafba02645d4b0e290/sigal/image.py#L137-L138

thinking that likewise the `context` dict can be modified by plugins so it should be returned and reassigned... without realizing that a dict is mutable and is passed by reference.

tests/test_encrypt.py
dbw9580: I copied those lines from `tests/test_extended_caching.py` and thought it looked good... I didn't realize a temp dir already takes care of the cleaning.

sigal/writer.py
dbw9580: Fixed in 2c4fdd6.

tests/test_encrypt.py
dbw9580: Removed in 2c4fdd6.

saimn: Indeed, actually `test_extended_caching` should use tmpdir instead, I will fix this in another commit.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+  specify a password.
+- ``ask_password``: Whether or not viewers are asked for the password to view 
+  the gallery. If set to ``False``, the password will be present in the HTML files
+  so the images are decrypted automatically. Defaults to ``False``.
+- ``gcm_tag``, ``kdf_salt``, ``kdf_iters``: Cryptographic parameters used when
+  encrypting the files. ``gcm_tag``, ``kdf_salt`` are meant to be randomly generated, 
+  ``kdf_iters`` defaults to 10000. Do not specify them in the config file unless 
+  you have good reasons to do so.
+- ``encrypt_symlinked_originals``: Force encrypting original images even if they 
+  are symlinked. If you don't know what it means, leave it to ``False``.
```
saimn, line 44: I think this should be True by default, and we could maybe even remove this setting (and send a warning that orig_link should not be used in this case). When using this plugin clearly the goal is to protect privacy by encrypting all images including the original ones.

sigal/plugins/encrypt/endec.py
```diff
@@ -0,0 +1,117 @@
...
+from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives import hashes
+from cryptography.exceptions import InvalidTag
+from typing import BinaryIO
+
+backend = default_backend()
+MAGIC_STRING = "_e_n_c_r_y_p_t_e_d_"
+
+def kdf_gen_key(password: bytes, salt:bytes, iters: int) -> bytes:
```
saimn, line 37: Could be better to accept str and move the various `.encode()` calls here.

sigal/plugins/encrypt/encrypt.py
dbw9580: Yes, but we need to also make sure people don't get their original pictures overwritten if they accidentally have both `keep_orig` and `orig_link` set to `True`, and have not made a backup. I think it's better we abort the encryption and send a warning, to make sure the user understands the situation, then it's up to them to decide. It's trivial to restart an aborted gallery build, while to recover the originals by decrypting them is not.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+# FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+# IN THE SOFTWARE.
+
+'''Plugin to protect gallery by encrypting image files using a password.
+
+Options:
+    encrypt_options = {
```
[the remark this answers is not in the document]
dbw9580: Fixed in 6c242b3.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+  are symlinked. If you don't know what it means, leave it to ``False``.
+
+Note: The plugin caches the cryptographic parameters (but not the password) after 
+the first build, so that incremental builds can share the same credentials. 
+DO NOT CHANGE THE PASSWORD OR OTHER CRYPTOGRAPHIC PARAMETERS ONCE A GALLERY IS 
+BUILT, or there will be inconsistency in encrypted files and viewers will not be able
+to see some of the images any more.
+'''
+
+import os, random, string, logging, pickle
```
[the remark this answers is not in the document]
dbw9580: Fixed in 6c242b3.

sigal/plugins/encrypt/endec.py
dbw9580: Fixed in 6c242b3.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+class Abort(Exception):
+    pass
+
+def gen_rand_string(length=16):
+    return "".join(random.SystemRandom().choices(string.ascii_letters + string.digits, k=length))
+
+def get_options(settings, cache):
+    if "encrypt_options" not in settings:
+        logging.error("Encrypt: no options in settings")
+        raise Abort
```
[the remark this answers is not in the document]
dbw9580: Fixed in 25f054b.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+    try:
+        options = cache["credentials"]
+    except KeyError:
+        options = settings["encrypt_options"]
+
+    table = str.maketrans({'"': r'\"', '\\': r'\\'})
+    if "password" not in settings["encrypt_options"] \
+        or len(settings["encrypt_options"]["password"]) == 0:
+        logger.error("Encrypt: no password provided")
+        raise Abort
```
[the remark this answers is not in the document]
dbw9580: Fixed in 25f054b.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+    copy(os.path.join(ASSETS_PATH, "sw.js"), settings["destination"], symlink=False, rellink=False)
+
+def inject_scripts(context):
+    try:
+        cache = load_cache(context['settings'])
+        context["encrypt_options"] = get_options(context['settings'], cache)
+    except Abort:
+        # we can't do anything useful without info in cache
+        # so just return the context unmodified
+        pass
```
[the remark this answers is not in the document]
dbw9580: Yes, I removed the try...except block so the `Abort` can fail the build in 25f054b.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,275 @@
...
+        logger.debug("encryption config: %s", config)
+        # make cache available from gallery object
+        # gallery.encryptCache = cache
+
+        logger.info("starting encryption")
+        copy_assets(settings)
+        encrypt_files(settings, config, cache, albums, gallery.progressbar_target)
+        save_cache(settings, cache)
+    except Abort:
+        pass
```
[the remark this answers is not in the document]
dbw9580: Indeed, no need to catch `Abort`s, changed in 25f054b.

dbw9580:
It occurs to me that the `zip_gallery` option will be meaningless if this plugin is enabled. The zip archive will contain encrypted images, and the viewers will not be able to decrypt them if they download the archive. We should add a note to the documentation about this.

sigal/plugins/encrypt/encrypt.py
saimn: Agreed! So I think you can remove this setting ?

dbw9580:
Remove from the doc?


2020年4月18日 10:23, 10:23，在 Simon Conseil <notifications@github.com> 已写:
@saimn commented on this pull request.

________________________________


In sigal/plugins/encrypt/encrypt.py<http://encrypt.py>:

 + 'kdf_iters': 10000, + 'encrypt_symlinked_originals': False + } + +-
``password``: The password used to encrypt the images on gallery build,
+ and decrypt them when viewers access the gallery. No default value.
You must + specify a password. +- ``ask_password``: Whether or not
viewers are asked for the password to view + the gallery. If set to
``False``, the password will be present in the HTML files + so the
images are decrypted automatically. Defaults to ``False``. +-
``gcm_tag``, ``kdf_salt``, ``kdf_iters``: Cryptographic parameters used
when + encrypting the files. ``gcm_tag``, ``kdf_salt`` are meant to be
randomly generated, + ``kdf_iters`` defaults to 10000. Do not specify
them in the config file unless + you have good reasons to do so. +-
``encrypt_symlinked_originals``: Force encrypting original images even
if they + are symlinked. If you don't know what it means, leave it to
``False``.

Agreed! So I think you can remove this setting ?

sigal/plugins/encrypt/encrypt.py
dbw9580: Remove it from the doc or the code?

saimn: Remove from the code, as you said above we should abort if `keep_orig` and `orig_link` are True, so I don't think we should keep this setting to force encryption. People should just choose between `orig_link` and using this plugin.

dbw9580: Good point. Changed in a3b0876.

sigal/plugins/encrypt/encrypt.py
dbw9580: On further investigation, I think I was wrong about caching exif. Unlike the sizes of the resized images and the thumbnails, the exif info is read from the source images, therefore it is always available: 

https://github.com/saimn/sigal/blob/8547f18e6be2f38485b6edaf89e59726466c4a0c/sigal/gallery.py#L200-L202

https://github.com/saimn/sigal/blob/8547f18e6be2f38485b6edaf89e59726466c4a0c/sigal/gallery.py#L216-L220

So it is not necessary to cache exif data. This issue is considered resolved.

saimn: Indeed, good point.

sigal/plugins/encrypt/encrypt.py
```diff
@@ -0,0 +1,271 @@
...
+            except Exception as e:
+                logger.error("Could not write to file %s: %s", filename, e)
+                return False
+    return True
+
+def copy_assets(settings):
+    theme_path = os.path.join(settings["destination"], 'static')
+    copy(os.path.join(ASSETS_PATH, "decrypt.js"), theme_path, symlink=False, rellink=False)
+    copy(os.path.join(ASSETS_PATH, "keycheck.txt"), theme_path, symlink=False, rellink=False)
+    copy(os.path.join(ASSETS_PATH, "sw.js"), settings["destination"], symlink=False, rellink=False)
```
saimn, line 261: Why `sw.js` is not in the static directory with the other static files ?

dbw9580: As per [Service Worker specification](https://www.w3.org/TR/service-workers/#path-restriction), the worker script has no access to resources outside its directory tree. So putting `sw.js` in `static` will break it.

saimn: Ok, thanks for the info!

saimn:
Great job, thanks @dbw9580 !

PR #387, opened 2020-04-11
Use natural sorting

dbw9580:
This PR uses `natsort` to replace `locale.strxfrm` to make sorting numbers in filenames and album names more human-friendly. `natsort` still handles locale-specific sorting through its internal use of `locale`.

This will add a new dependency on `natsort`.


saimn:
Seems sensible, thanks @dbw9580.

PR #391, opened 2020-05-22
wip: add nginx mod_zip output feature

faudebert:
This a draft pull-request (i.e. no test, no doc yet). I would be happy to get feedback to know if maintainers are interested by such a feature. If so, I will take time to make it cleaner including docs, samples and tests.

It's great being able to provide an album download link thanks to `zip_gallery` plugin. Nonetheless, on my Sigal instance, it is a seldomly used feature. Thus, having many zip archives on my hard drive which duplicates all my original pictures is a waste of disk storage.

The idea of this pull-request is to make Sigal generate NGINX [`mod_zip`](https://www.nginx.com/resources/wiki/modules/zip/) "manifest" instead of real zip archives. Then, on user request, a properly configured NGINX will take care of dynamically generating the zip archive using the "manifest".

P.S.: I've currently re-used `zip_gallery` since it's very similar job (only the output result change), not sure it's the best approach though.

### DIFF
--- a/sigal/plugins/zip_gallery.py
+++ b/sigal/plugins/zip_gallery.py
@@ -31,10 +31,14 @@ present, then make a ZIP archive with all media files.
 See :ref:`compatibility with the encrypt plugin <compatibility-with-encrypt>`.
 """
 
+import io
 import logging
 import os
 import zipfile
-from os.path import isfile, join, splitext
+import zlib
+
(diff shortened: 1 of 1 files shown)


saimn:
Seems very interesting ! :+1: 
Reusing `zip_gallery` seems to make sense.

PR #398, opened 2020-07-26
theme photoswipe: fix video playback and fullscreen (on big screen)

luv:
This PR fixes photoswipe and video playback:

- most importantly removes "padding: 2% 0 15% 0;" from ".video__container video" and " padding-bottom: 56.25%;" from ".video__container"
- Now the videos fill the screen properly (max-height and max-weight 100%), if the video is smaller than display size it is centered both vertically and horiziontally.
- add a play button in the gallery overview so the users know it's a video

I tested the above changes only in Firefox on Linux desktop.

I will fix photoswipe on mobile later (most importantly, the video playback should stop when you swipe away and it should be easier to swipe when a video is shown)


luv:
Converting to a draft as I'm still working on fixing photoswipe on mobile, and I'd like to make it easily configurable to render a dark theme as well.

luv:
Quick video to demonstrate now it works correctly even when the video is bigger/smaller than the window size and even for weird aspect ratios.
output

luv:
fixed video playback: on close/go to another item, the video is paused now

luv:
Next step: (mobile) it should be easier to swipe to next/previous item  when a video is playing

luv:
Now a swipe to next item pauses the video on android as well. It's pretty annoying when you change to next/picture and the previous video with sound is still playing.

Embedded video on iphone behaves somehow differently (handled by the iOS internal player I assume because all browsers handle it the same) and photoswipe has been developed and tested only on iOS that's why it has this problem on Android.

luv:
Swiping over video now works.

next: (configurable) dark theme and, if possible, configurable "playsinline" video option for iphone, less margin from top if gallery does not have a title

luv:
Added 2 new config optins (to be used in sigal.conf.py):

- photoswipe_notitle
- photoswipe_darktheme

The final step is to add support for configurable "playsinline" option for videos to play "inline" on iOS devices and test through all the browser/devices I have access to (unfortunately I wont be able to test on android now).

luv:
All functionality now in, tested on chrome and firefox on linux and safari and firefox on iOS with and without "playsinline".

Three new photoswipe-specific configuration options now available:

- photoswipe_notitle
- photoswipe_darktheme
- photoswipe_video_playsinline

Ready for review.

To demonstrate I spinned up a temporary cloud instance (ie. the urls below will stop working in few days):

- http://45.77.248.75/d  - dark theme without title with "playsinline"
- http://45.77.248.75/l  - default theme, default title, no "playsinline"
- http://45.77.248.75/m  - current master

(not that the first video does not work on iOS because iOS is very picky about video and audio format however the second video is formatted for iOS for demonstration purposes)

You can notice that following issues have been fixed:
- better scaling of video on desktop and android (on iOS without "playsinline" the strange big bottom margin makes sense but even with proper scaling it works very well), lower resolution videos are still not upscaled
- video is paused when not focused (the item is not destroyed so it's still cached, it's just paused)
- swiping works even when video is focused (on master this is an issue on firefox on android and possibly other browsers too, fixed in the PR by rather ugly hack, but now swiping and controls work on all major browsers)
- configurable darktheme, "playsinline" for iOS and "notitle" option
- videos are "marked" with a "play" icon so the user know he/she is clicking on a video (I have created the icon in gimp so there are no copyright issues and .xcf file is also attached)

I understand the swipearea hack is very controversial, maybe we can make it firefox-only or configurable or scrap it altogether.

saimn:
Thanks for the work @luv , it looks very good.
Could you add the new options to the template settings file (`templates/sigal.conf.py`) ?

Also my main worry is that you are modifying photoswipe itself ... photoswipe seems to be unmaintained lately which is a bit sad, but if there is a new version in the future it will be painful to update. And if there is no new version, I guess at some point I would just remove the photoswipe theme, there are other alternatives.

luv:
Thanks for your positive feedback. Unfortunatelly I don't have access to an android phone at the moment so I have not verified everything works as expected on Android.

However except for the "swipe area hack" the changes to the theme are pretty straight forward.

It looks like that hack is needed only for firefox and because almost noone is using firefox on mobile it can be removed. I just want to make sure it works on chrome on android before I remove that swipe area workaround. I don't think it's worth keeping such a workaround in only because of a problem in firefox on android (can't swipe over video, works on chrome on linux as well in iOS browsers).

I will check with chrome on android later (but this may take few weeks before I have a chance to properly test on android) and add the settings to `templates/sigal.conf.py`.

I understand this PR is not really for sigal but for photoswipe, I didn't know photoswipe is unmaintained and is considered for removal, I have to say that except for the problems addressed in this PR sigal+photoswipe works perfectly for me for self-hosted photo and video galleries that work great on both mobile and desktop.

If photoswipe is abandoned I could fork and maintain the theme.

Are there any alternatives to photoswipe which work for both photo and video and on both desktop and mobile you can recommend?

fidergo-stephane-gourichon:
Hi. Look likre the hack is this line: https://github.com/saimn/sigal/pull/398/commits/7c69a43a5995b047db72aedb7688c6c7b6cbb70a#diff-52dca0b6c4fd6b9cc046c9ff592e1040R45 where `playsinline` gets surrounded by double braces.

> It looks like that hack is needed only for firefox and because almost noone is using firefox on mobile it can be removed. 

Wow! I know Firefox user shares have been declining and the situation is worrying, but it's not dead yet and producing new improved versions regularly.

Can you clarify what happens without this hack? (I agree with @saimn about being worried about modifying photoswipe itself.)

luv:
Hi. I agree that modifying a bundled theme is pretty much unmaintainable. If the changes can be done upstream, great. If not I'm not sure how to go about this (if there is a better/different theme that can be used instead of photoswipe I'm happy to drop this and use the new theme instead myself). I could potentially maintain the theme myself (with minimal changes) and because themes in sigal are nicely pluggable, when it's dropped/replaced anyone could just get the (fixed) theme from "my github". That's more or less what I do now, I just keep the copy of my fixed theme in my gallery source directory and set theme to './photoswipe' and everything works nicely.

Don't worry, I still use firefox myself on my linux desktop and even without the workaround everything works fine on desktop firefox, firefox on android is very niche yet I would still like to support it (afterall I used firefox on android myself and that's how I discovered the problem), the "swipe area hack" I talk about is actually very huge potentially problematic workaround when **a div is drawn on top of the video area** and **almost** covers the video area (and every tap/click is processed as pause/play toggle so it works "transparently" from user point of view) this is done because firefox video player "eats" all the events so swipe over video can't be implemented. To implement this properly is actually a rather big fragile workaround because of all the sizing, resizing and positioning that's why I think it's not worth it even though I used firefox on android myself. 

The workaround is this commit https://github.com/saimn/sigal/pull/398/commits/1b9751b427f4b96c30e4205e13b693f358788d13 (and subsequent fixup https://github.com/saimn/sigal/pull/398/commits/7ebc7876683974cdd1b13460b125b658f6990380 so the "swipe area div" does not cover top controls of iOS video player).

saimn:
> I understand this PR is not really for sigal but for photoswipe, I didn't know photoswipe is unmaintained and is considered for removal

I didn't know either but had a look since indeed your PR would better fit in Photoswipe itself ;).
https://github.com/dimsemenov/PhotoSwipe : Last commit from Jan, 2019, and many issues and PRs 
It seems that someone is starting a fork : https://github.com/andi34/PhotoSwipe

> If photoswipe is abandoned I could fork and maintain the theme.

Yes, at least I think it would help to have the modifications on Photoswipe done in a separate repository, then sigal wold just "depend" on this fork. This would make it easier in the future if Photoswipe finds new maintainers, and a well maintained fork appear. From my point of view I just want to be able to update it easily, integrating a new version in this repo when it is available. And if nothing happen and we find a better alternative, maybe at some point I will remove the theme, but there is no urgency at all. For now it's better if we can find a way to integrate fixes without too much pain.

> Are there any alternatives to photoswipe which work for both photo and video and on both desktop and mobile you can recommend?

Not really, I haven't been looking at this recently.
https://fengyuanchen.github.io/viewerjs/ seems nice but does not support videos.

fidergo-stephane-gourichon:
> It seems that someone is starting a fork : https://github.com/andi34/PhotoSwipe

Indeed this one shows a significant number of commits beyond the parent's master, and even new releases 4.2.0 and 4.3.0 .

There seem to be around 1000 forks on https://github.com/andi34/PhotoSwipe/network/members

I know most forks are "shallow", but with 1000 forks, there may be others. Even https://github.com/andi34/PhotoSwipe/network fails to load for me.
 
How did you find andi34? Generally, how can you find really active forks?

saimn:
I had a look at the recent issues and PRs, and the fork's author left some comments there.

saimn:
@luv - Since https://github.com/andi34/PhotoSwipe seems actively developed/maintained, I think I will switch to it (#415). Most of your changes here are still valid, and much appreciated. Do you think you could remove the changes to photoswipe itself, and maybe propose them on the fork ?

luv:
Moving PhotoSwipe to the active fork sounds great, thanks for putting in the work to keep sigal well maintained.

Unfortunately in the coming weeks I won't have time to clean my changes and rebase them on top of the fork. However I keep this task on my list and I will try to get my changes merged into that fork later on (but it won't be this year).

saimn:
Ok, thanks for the update, and no worries for the time it takes, it's also hard for me to find time to work on sigal ;).

Issue #407, opened 2020-11-01
img_format creates files with wrong suffix


hjbaader:
Hi,

I used to have only JPEG images in my archive, now I have some TIFF Images too. So I set img_format to JPEG. Indeed it creates JPEG files from the TIFF, but it keeps the suffix .tif or .tiff rather changing it to .jpg.

I use Sigal 2.1.1 (just upgraded).

Thanks,
Hans-Joachim

saimn:
Right, this was implemented in #360, but indeed it changes only the format but not the extension.

hjbaader:
In the meantime I have patched the code and made it work for me. I'll try to make a patch against the current version.

saimn:
Would be great to have a PR fixing this, thanks :)

PR #412, opened 2020-12-03
Make sure files have the suffix .jpg when their format is JPEG

hjbaader:
This is my solution to #407. It works for me, but it may be incomplete or unclean.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -140,11 +140,11 @@ class Media:
                 s = self.settings
                 if self.type == 'image':
                     image.generate_thumbnail(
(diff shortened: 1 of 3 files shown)


saimn:
Thanks. I need to have a careful look at this, since - as you noticed - the media class does not distinguish the input and output extensions. I also have a local branch with commits, but this is a bit of a mess and I didn't have time to make progress.

saimn:
@hjbaader - I finally found some time to look at this and ended up doing a lot more changes to better separate inputs and outputs:  #421. Thanks anyway for your contribution!

Issue #413, opened 2020-12-03
Slow and/or bad conversion of RAW images to JPEG


hjbaader:
I plan to save only RAW images rather than the camera generated JPEG in the future. Making such images into gallery images might work with some formats, but it is slow because a giant TIFF file or whatever ist generated by Pillow. For my images in CR2 format, it produces mostly trash.

However the CR2 files contain a full-sized JPEG preview of the image in acceptable quality. So we could create a gallery image by simply extracting this JPEG and resizing it. I have created a patch which works for me but needs to be improved because it hard-codes everything. I'll make a pull request for checking it out.

saimn:
For the record, from the discussion in #414:

> I would prefer to avoid dependency on external tools though, and I think there is a pretty good Python alternative: https://github.com/letmaik/rawpy
rawpy is based on libraw so it supports a wide range of raw formats (CRW/CR2, NEF, RAF, DNG, and others), and the README has an example of how to extract the preview image.

PR #414, opened 2020-12-03
Use embedded preview from raw image rather than converting the raw

hjbaader:
This is a patch for #413. Not really applicable, but perhaps some would like to have it as an option.

### DIFF
--- a/sigal/image.py
+++ b/sigal/image.py
@@ -36,6 +36,7 @@ import sys
 import warnings
 from copy import deepcopy
 from datetime import datetime
+from tempfile import TemporaryDirectory
 
 from PIL import Image as PILImage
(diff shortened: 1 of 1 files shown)


saimn:
Something like this would be very useful indeed! I would prefer to avoid dependency on external tools though, and I think there is a pretty good Python alternative: https://github.com/letmaik/rawpy
rawpy is based on libraw so it supports a wide range of raw formats (CRW/CR2, NEF, RAF, DNG, and others), and the README has an example of how to extract the preview image.

hjbaader:
I'd like to retire this PR. I have restructured my archive so that the public part only contains JPEG images and all raw images are in my internal archive. So I don't use it anymore.

PR #420, opened 2021-01-30
Additional video options

ksfeldman:
Adding in the ability to do 2-pass encoding
Allowing user to disable video resizing
Allowing user to force a video encode even within the same codec


sigal/templates/sigal.conf.py
```diff
@@ -171,14 +171,35 @@
...
 # video_size = (480, 360)
 
+# If the desired video extension and filename are the same, the video will
+# not be converted. If a transcode to different quality is required,
+# set this to True to force convert it. False by default.
+# video_always_convert = False
+
+# Set this to false if no resizing is desired on the video. This overrides
+# the video_size option.
+# video_resize = True
```
saimn, line 201: Maybe allow `video_size = None` to disable resizing, which would remove one configuration option ?

sigal/video.py
```diff
@@ -78,30 +78,36 @@ def video_size(source, converter='ffmpeg'):
...
     logger = logging.getLogger(__name__)
 
     # Don't transcode if source is in the required format and
     # has fitting datedimensions, copy instead.
-    converter = settings['video_converter']
     w_src, h_src = video_size(source, converter=converter)
-    w_dst, h_dst = settings['video_size']
+    w_dst, h_dst = output_size
     logger.debug('Video size: %i, %i -> %i, %i', w_src, h_src, w_dst, h_dst)
+    return {'src': (w_src, h_src), 'dst': (w_dst, h_dst)}
```
saimn, line 96: Why not putting this directly in `get_resize_options` to avoid the dict packing/unpacking ? I think it's the only place where it is used.

sigal/video.py
```diff
@@ -112,25 +118,78 @@ def generate_video(source, outname, settings, options=None):
...
+        shutil.copy(source, outname)
+        return
+
+    final_pass_options = _get_empty_if_none_else_variable(options) + resize_opt
+    if second_pass_options:
+        generate_video_pass(converter, source, final_pass_options)
+        final_second_pass_options = _get_empty_if_none_else_variable(
+                    second_pass_options) + resize_opt
+        generate_video_pass(converter, source,
+                            final_second_pass_options, outname)
```
saimn, line 185: So when doing 2-pass encoding the idea is that ffmpeg will create a temp file in the first pass and find it when doing the second pass right ?

ksfeldman: That is correct, yes. I don't believe there is a way to change the filename from the default "ffmpeg2pass-0.log"

ksfeldman:
Thanks for the recommendations, both have been pushed. Would you like me to squash this commit as well?

saimn:
Oups sorry I forgot to merge your PR before mine (#421), so now you get some conflicts :(
It should be easy to fix, just 2 lines in `process_video`, but I can do it if you prefer.
Also thanks for the change, looks good.

ksfeldman:
No problem, conflict should be resolved. Merge whenever! Thanks

saimn:
Thanks!

PR #434, opened 2021-06-28
plugin to handle non-image or video files

### LINKED ISSUE #432: Allow non-image files in the gallery
I have a need for adding files that aren't images or videos to a gallery, specifically for pdfs and other media assets that may not be preview-able in a browser.  I'm wondering what the best way to go about this would be (I'm willing to make the modifications myself).

One option would be to make a plugin to handle "arbitrary" file types, and add a signal to hook into the file processing loop in `Album.__init__()`.  This seems better than making a global class, as it should affect fewer people.

Thoughts @saimn?

dsschult:
This PR fixes #432 with a plugin + a few other changes to handle non-image or video files.  

The default thumbnail is the file extension.  Here's an example for a pdf, which is simplistic at best:

nonmedia_thumbnail_pdf

What I have now works for me, but I'm also using a custom theme that treats these specially (no preview, just a link to the file).  I haven't spent much time looking at what happens in the regular themes.  Let me know if I should put more effort into that.


dsschult:
I had to make the font type selection optional to get the bot to pass, since Actions doesn't have a full install of fonts. The default font is fairly poor though.

`FreeMono.ttf` is a nice font that is likely found on Linux systems (and also what the Pillow example uses), but not sure if that should be recommended anywhere.  It's probably best to avoid putting a selection algorithm in code, though maybe adding a paragraph in the documentation explaining the thumbnail generation is better?

saimn:
I will have a more detailed look and play with it later, but looks really great !

sigal/plugins/nonmedia_files.py
```diff
@@ -0,0 +1,184 @@
...
+    """Gather all informations on a non-media file."""
+
+    type = 'nonmedia'
+
+    def __init__(self, filename, path, settings):
+        super().__init__(filename, path, settings)
+        self.thumb_name = os.path.splitext(self.thumb_name)[0] + '.jpg'
+        self.date = self._get_file_date()
+        self.mime = get_mime(self.src_ext)
+        logger.debug('mime type %s', self.mime)
```
saimn, line 77: `self.mime` doesn't seem to be used ? Or do you think it's useful to have it for the templates ?

sigal/plugins/nonmedia_files.py
```diff
@@ -0,0 +1,184 @@
...
+                media.thumb_path,
+                settings['thumb_size'],
+                options=settings['jpg_options'],
+                **kwargs
+            )
+        except Exception:
+            if logger.getEffectiveLevel() == logging.DEBUG:
+                raise
+            else:
+                return Status.FAILURE
```
saimn, line 162: Would be better to return a `Status.SUCCESS` at the end (it works without but for consistency with other processors).

saimn (commented):
Thanks for the great work @dsschult, looks perfect, just a small question.

sigal/plugins/nonmedia_files.py
dsschult: I added it for the templates.  Not sure if it will be used, but it was simple to add.

sigal/plugins/nonmedia_files.py
dsschult: Yes, you're right.  I'll change that.

saimn:
Sorry wrong button :D. Thanks @dsschult !

PR #435, opened 2021-06-28
allow video_size = None in settings.py

dsschult:
A small fix to allow `video_size = None`, which is the way to disable video resizing.


saimn:
Thanks!

Issue #408, opened 2020-11-05
Album cannot contain both subfolders and images


thomasdn:
Hi,

When running sigal on a folder structure as the one below, sigal complains with this warning:

> WARNING: Album X contains sub-albums and images. Please move images to their own sub-album. Images in album X will not be visible.                                                                                                                       

This results in all the images directly in the album not being visible. Only those in subfolders.
I think this is unfortunate as it is quite useful to have subfolders within an album. Especially if it is a big album. For example, when I have pictures from a vacation, I might want to put all the videos in a separate subfolder called videos. Also, I sometimes take pictures of food. Most people are uninterested in those pictures. So I put all the food related pictures in a subfolder.

The folder structure that produces the above problem is:

    pictures/
    pictures/vacation-pictures/
    pictures/vacation-pictures/img_001.jpg
    pictures/vacation-pictures/img_002.jpg
    pictures/vacation-pictures/img_003.jpg
    pictures/vacation-pictures/food/
    pictures/vacation-pictures/food/beautiful-sandwich.jpg
    pictures/vacation-pictures/videos/
    pictures/vacation-pictures/videos/video1.mp4

In this case, the following pictures will be omitted form the gallery:

    pictures/vacation-pictures/img_001.jpg
    pictures/vacation-pictures/img_002.jpg
    pictures/vacation-pictures/img_003.jpg

fidergo-stephane-gourichon:
Hi! Another user here. I also noticed this and it feels like it could be improved.

I had the opportunity to visit sigal source code in the past (I think the design is very well thought out) and remember two templates.

@saimn is this limitation "only" caused by having two templates, one for a page listing sub-albums, one for a page listing photos?

Could it be solved by having one template containing both parts, which would be activated or not depending on whether the current directory contains photos and/or contains subfolders?

Thanks!

thomasdn:
I think the right way to present an album containing both images and sub albums is to present it with all the subalbums/folders first followed by all the images. 
Maybe this could be done by creating a parent template that includes a subtemplate for presenting subalbums followed by a subtemplate for presenting images.

Since the templates are based on jinja2, it should be possible to use template inheritance to solve this, right?

saimn:
> is this limitation "only" caused by having two templates, one for a page listing sub-albums, one for a page listing photos?

[...]

PR #438, opened 2021-07-11
Added common subtemplate for album items to be included from album.ht…

thomasdn:
…ml and album_list.html


sigal/themes/galleria/templates/album_items.html
```diff
@@ -0,0 +1,231 @@
...
+        </div>
+        {% endif %}
+
+        {% if album.description %}
+        <div id="description">
+          {{ album.description }}
+        </div>
+        {% endif %}
+
+{% block late_js %}
```
saimn, line 25: Just to be sure, is this loaded in the correct block (`late_js`) even if it was included in another block (`content`) ?

saimn: Also one drawback of this is that the javascript includes and script will always be loaded in the album list pages pages, even of not needed. I  guess we could prevent that by checking if `album.medias` is not None.

thomasdn: > Just to be sure, is this loaded in the correct block (`late_js`) even if it was included in another block (`content`) ?

I think so. At least it seems to work correctly in the tests I performed in Chrome and Firefox.

thomasdn: > Also one drawback of this is that the javascript includes and script will always be loaded in the album list pages pages, even of not needed. I guess we could prevent that by checking if `album.medias` is not None.

I agree that this could be optimized to not load the javascript when there is no need. Checking album.medias seems like a proper way of doing that.

saimn:
Merged manually in c40832c7fef4dce24bdfff725d362f66aae2ef72 to remove the extra merge commit. Thanks @thomasdn.

saimn:
Ref #408

PR #440, opened 2021-07-20
Sqlite cache for metadata lookup

dsschult:
Add an optional sqlite cache for metadata lookup, instead of opening md or image files (for exif data). This greatly speeds up rebuilding for large libraries.  The cache does check the file modification time, so will allow updates.

As an example, I have ~50k files in a library, and this cache reduces the time by half when running updates.

### DIFF
--- a/sigal/__init__.py
+++ b/sigal/__init__.py
@@ -108,13 +108,15 @@ def init(path):
 )
 @option('--title', help="Title of the gallery (overrides the title setting.")
 @option('-n', '--ncpu', help="Number of cpu to use (default: all)")
+@option('--cache', help="Cache file path")
 def build(
-    source, destination, debug, verbose, quiet, force, config, theme, title, ncpu
+    source, destination, debug, verbose, quiet, force, config, theme, title,
+    ncpu, cache
 ):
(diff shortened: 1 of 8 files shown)


saimn:
Thanks @dsschult, but there is already a plugin for this (https://sigal.saimon.org/en/latest/plugins.html#module-sigal.plugins.extended_caching). It doesn't cache the markdown files, but the gain should be small compared to parsing EXIF metadata. Could you have a look if you missed it, and maybe see if it can be improved (if there is a gain) ? Also I would prefer keeping the core as simple as possible, so the plugin approach seems better.

saimn:
That's surprising, this plugin is caching the EXIF metadata for each image. I don't understand your sentence "The metadata exif is completely separate." ?

dsschult:
Sorry, looking over the code in more detail, they are connected, in that the exif data is loaded in the `_get_metadata()` function and then used in `exif()`.  But it doesn't go in reverse, so even though the `extended_caching` plugin overrides `exif`, the exif data is always read from file in `_get_metadata()`.

dsschult:
Now using #443 instead.

PR #443, opened 2021-08-07
Extend caching plugin to handle markdown and iptc metadata

dsschult:
Turn markdown metadata in `gallery.py` into a property so it can be overridden.
Extend the caching plugin to handle markdown and iptc metadata (previously only handled exif).
Invalidate cache when a file timestamp changes, so we always get the latest data.
Cache file modification time lookups within a build to reduce load on filesystem.

Supersedes #440.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -188,33 +188,33 @@ class Media:
     @cached_property
     def description(self):
         """Description extracted from the Markdown <imagename>.md file."""
-        return self.raw_metadata.get('description', '')
+        return self.markdown_metadata.get('description', '')
 
     @cached_property
     def title(self):
         """Title extracted from the metadata, or defaults to the filename."""
-        title = self.raw_metadata.get('title', '')
+        title = self.markdown_metadata.get('title', '')
         return title if title else self.basename
 
     @cached_property
     def meta(self):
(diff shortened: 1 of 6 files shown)


corsmith:
I have an album with 150K pictures and 4K videos and use the extended_caching plugin.  The 2.2 release takes approximately 18 minutes to rebuild the gallery with the plugin enabled (assuming no thumbnails to generate) and the cache file is approximately 23MB.  Using the code in this PR it takes less than 5 minutes to rebuild the gallery with a cache size of 1.03GB (43x the previous size!).  The pickle.load call in my setup alone takes 60 seconds of CPU time!  While that is a little crazy the net effect is a 60% reduction in wall time to complete a full rebuild even with the previous extended_caching in 2.2. It is faster to parse the single file than to read 154K files for metadata across the filesystem as currently done in gallery.py -> class Media -> init -> self._get_metadata()

saimn:
Sorry it seems I forgot to review this PR. I just had a look and it looks very good. 
@dsschult - could you rebase it on main to make sure tests still run ? 
If I had one comment about the code, maybe `raw_metadata` could be renamed to something like `markdown_metadata` or `mkd_metadata` ?

dsschult:
@saimn should be good now

saimn:
Excellent, thanks for the quick update !

PR #446, opened 2021-08-26
Add original size information for 'Download ZIP' link in album

vsg005:
For all themes (colorbox, galleria, photoswipe), modified "Download ZIP" link to "Download ZIP with original files" if zip archive was created with original files.


saimn:
Thanks !

Issue #451, opened 2022-01-07
`filters.py` fails to loads since release 2.2


vszakats:
2.2 introduced https://github.com/saimn/sigal/commit/ceab382e0252c4d7b25f8c2905b64628a9e15ba6, and with the new custom filter loading method, a previously working `filters.py` now results in this error:
```
Traceback (most recent call last):
  File "/.../.venv/bin/sigal", line 8, in <module>
    sys.exit(main())
  File "/.../.venv/lib/python3.9/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/.../.venv/lib/python3.9/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)                                        
  File "/.../.venv/lib/python3.9/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))                     
  File "/.../.venv/lib/python3.9/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)                              
  File "/.../.venv/lib/python3.9/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)                                          
  File "/.../.venv/lib/python3.9/site-packages/sigal/__init__.py", line 157, in build
    gal.build(force=force)                                                      
  File "/.../.venv/lib/python3.9/site-packages/sigal/gallery.py", line 781, in build
    album_writer = AlbumPageWriter(self.settings,                
  File "/.../.venv/lib/python3.9/site-packages/sigal/writer.py", line 83, in __init__
    mod = importlib.import_module(filters_py)                    
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)  
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import  
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named '/.../filters'
```

[...]

PR #452, opened 2022-01-07
fix loading filters.py using importlib

vszakats:
Regression from: https://github.com/saimn/sigal/commit/ceab382e0252c4d7b25f8c2905b64628a9e15ba6
Ref: https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

Fixes https://github.com/saimn/sigal/issues/451

### DIFF
--- a/sigal/writer.py
+++ b/sigal/writer.py
@@ -93,7 +93,10 @@ class AbstractWriter:
         # handle optional filters.py
         filters_py = os.path.join(self.theme, "filters.py")
         if os.path.exists(filters_py):
-            mod = importlib.import_module(filters_py)
+            module_spec = importlib.util.spec_from_file_location('filters', filters_py)
+            mod = importlib.util.module_from_spec(module_spec)
+            sys.modules['filters'] = mod
+            module_spec.loader.exec_module(mod)
             for name in dir(mod):
                 if isinstance(getattr(mod, name), types.FunctionType):
                     env.filters[name] = getattr(mod, name)


saimn:
Thanks for the fix! Do you think you could try to add a unit test for this ? Would be great to make sure this feature is not broken again.

vszakats:
That would definitely be a good idea. But, after an hour trying to implement it, I'm out of ideas how to do it. The sample project under tests uses a built-in theme and the filter is loaded from the theme root with no way to override (but even when copied there, it's not loaded). Do you have some suggestions what would be a good way to implement this?

`filters.py`:
```python
def test_filter():
    return "filter 'test_filter' is working"
```

```diff
--- a/tests/sample/pictures/index.md
+++ b/tests/sample/pictures/index.md
@@ -15,3 +15,7 @@ in urna ultrices accumsan. Donec sed odio eros. Donec viverra mi quis quam
 pulvinar at malesuada arcu rhoncus. Cum sociis natoque penatibus et magnis dis
 parturient montes, nascetur ridiculus mus. In rutrum accumsan ultricies.
 Mauris vitae nisi at sem facilisis semper ac in est.
+
+## Filter test
+
+{{ test_filter }}
```

saimn:
Yeah I agree it's not the easiest to test. I will have a look when I can but my time for sigal is also limited so no worries, and let's at least merge the fix. Thanks for contributing !

PR #455, opened 2022-01-22
colorbox theme: avoiding holes due to non image/video files present in the source directory - fix #454

### LINKED ISSUE #454: colorbox theme: non image/video files create holes in the grid
In [album.html @colorbox theme](https://github.com/saimn/sigal/blob/main/sigal/themes/colorbox/templates/album.html#L19), files are only included in the HTML page if `media.type == "image"` or `media.type == "video"`, leaving holes in the grid.

Example with a hole due to `sigal.conf.py` being present in the source directory:
WebWittyComeBack

PR incoming!

Lucas-C:
Fix #454.

### DIFF
--- a/sigal/themes/colorbox/templates/album.html
+++ b/sigal/themes/colorbox/templates/album.html
@@ -16,7 +16,7 @@
 
   {% include 'map.html' %}
   <div id="gallery">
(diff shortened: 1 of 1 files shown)


saimn:
Thanks!

PR #456, opened 2022-01-29
Update custom theme templates, document default theme

japareaggae:
The documentation referenced an "index.html" template which appears to
be no longer used. Update the documentation so that it talks about the
correct templates, album.html and album_list.html.

Also, add documentation about the default theme, talking about what it
contains and what it doesn't.


saimn:
Thanks!

PR #468, opened 2022-06-09
List of sort-properties as fallback instead of zero-length string

tudacs:
This PR contains code to be able to replace the scalar 'albums_sort_attr' by a list of properties to take into account as keys for sorting.

Thus you may have multiple 'fallback' options for sorting and prevent unpredictable outcome of natsort, which previously was fed with a zero-length string for all items who did not offer the specified property. E.g. `albums_sort_attr = ['meta.sortname','filename']` would take the filename for those items who do not have `meta.sortname` available.

### DIFF
--- a/sigal/gallery.py
+++ b/sigal/gallery.py
@@ -488,18 +488,26 @@ class Album:
                     reverse = False
 
             root_path = self.path if self.path != '.' else ''
-            if albums_sort_attr.startswith("meta."):
-                meta_key = albums_sort_attr.split(".", 1)[1]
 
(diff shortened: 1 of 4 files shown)


saimn:
Sounds useful, thanks. Could you add some tests for this ?

tudacs:
Sure! Thanks for the reminder. While developing I had some (of the already included) tests failing, I guess that's why I forgot to add tests for the added features - sorry.

saimn:
Thanks!

PR #469, opened 2022-06-10
Another plugin: Titleregexp - modify titles of albums with regular expressions

tudacs:
This PR adds a plugin called 'titleregexp' which modifies titles of albums by configured regular expressions and simple search-and-replace tuples.

### DIFF
--- a/sigal/plugins/titleregexp.py
+++ b/sigal/plugins/titleregexp.py
@@ -19,23 +19,37 @@
 # IN THE SOFTWARE.
 
 """ This plugin modifies titles of galleries by using regular-expressions and
-simple character replacements.
+simple string or character replacements. It is acting in two phases: First, all
+regular-expression-based modifications are carried out, second, string/character
+replacements are done.
+
+The first phase may be interrupted if a match occurs by the 'break'-setting
+individually per regular-expression part. Additionally, if a match occurs,
+string/character replacements may be added.
+
+The second phase is done even if the first phase had been interrupted.
 
 Settings:
 
 - ``titleregexp`` with the following keys:
-    - ``regexp``, which is an array of dicts with 'search', 'replace' and 
-        'count' keys
+    - ``regexp``, which is an array of dicts with 'search', 'replace', 'break'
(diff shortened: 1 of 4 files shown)


sigal/plugins/titleregexp.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+import logging
+import os
+import re
+
+from sigal import signals
+
+logger = logging.getLogger(__name__)
+cfg = {}
```
saimn, line 52: Define `cfg` here seems useless.

sigal/plugins/titleregexp.py
```diff
@@ -0,0 +1,85 @@
...
+simple character replacements.
+
+Settings:
+
+- ``titleregexp`` with the following keys:
+    - ``regexp``, which is an array of dicts with 'search', 'replace' and 
+        'count' keys
+    - ``substitute``, which is an array of 2-element-arrays, of which the
+        second element denotes the replacement of occurences of the first 
+        element
```
saimn, line 31: Could you explain a bit more what the regex is doing ? (I see it's using `re.subn` but it's not clear from the docs)

sigal/plugins/titleregexp.py
```diff
@@ -0,0 +1,85 @@
...
+            for s in r.get('substitute', []) :
+                album.title = album.title.replace(s[0],s[1])
+            if r.get('break','') != '' :
+                break
+
+    for r in cfg.get('substitute', []) :
+        album.title = album.title.replace(r[0],r[1])
+
+    if total > 0:
+        logger.info("Fixing title to '%s'", album.title)
```
saimn, line 78: Maybe log the original title as well, to make it clear which one has been changed.

saimn (commented):
A few comments below. Also please add your plugin to `docs/plugins.rst`, thanks.

tudacs:
Thank you for looking into this, your remarks are appreciated :). I added the documentation as suggested and tried to be as clear as possible. If the docs are too technical please let me know and I'll try to adapt it.

saimn:
Looks good, thanks.

PR #471, opened 2022-06-27
Make theme dir writable after copying to it

### LINKED ISSUE #458: Gallery creation fails if THEMES_PATH is readonly
I'm seeing this on NixOS, where installed software is all stored on a read-only filesystem, but I suspect it can be reproduced on any system where the contents of THEMES_PATH have mode `a-w`.

I've reproduced it on both 2.2 and HEAD; the specific error is different (since HEAD no longer has a custom `copytree` implementation), but the symptom is that when running `sigal build`, it processes all the images fine and then dies with `permission denied` when attempting to copy the theme files into place.

As best I can figure, what happens is this:
- it gets to [this point](https://github.com/saimn/sigal/blob/246aed53ff2d29d680cc81929e59f19023e463bb/sigal/writer.py#L103) in `writer.py` and copies `$THEMES_PATH/default/` into `static/`, **preserving permissions**. This means that if THEMES_PATH is readonly, the copy will inherit that.
- it then tries to copy the files for the specific theme into place, and can't because everything in `static/` has been marked readonly.

Note that THEMES_PATH doesn't need to be writeable *by the user invoking sigal* -- e.g. it's ok if it's installed somewhere in `/usr/share` and has permissions `755:root:root` -- but it has to have that write bit set so it'll be carried over by the `copytree`.

I don't know what's a good solution to this. `copytree` lets you specify a custom copy function, so you could pass it something other than `copy2` that doesn't preserve permissions or calls `os.chmod` afterwards to make sure the files are writeable, but that won't stop it from copying directory permissions over. As a quick and dirty hack, I've patched it to let me specify `THEMES_PATH` externally (via the `SIGAL_THEMES_PATH` environment variable) and then copied the entire themes dir into `~/.cache/sigal/` and made it writeable, but I don't think that's a production-ready fix.

amesgen:
Closes #458.

Not really elegant, but it does the job for me.

### DIFF
--- a/sigal/writer.py
+++ b/sigal/writer.py
@@ -24,6 +24,7 @@ import importlib
 import logging
 import os
 import shutil
+import stat
 import sys
 import types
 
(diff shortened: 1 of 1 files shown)


saimn:
Thanks!

PR #480, opened 2022-08-23
This plugin add basic raw support.

poucz:
Plugin make temp directory which is use for generate jpeg from raw file (jpeg is camera-generated thumbnail).
For this dcraw program is used.

### DIFF
--- a/src/sigal/plugins/lrp.py
+++ b/src/sigal/plugins/lrp.py
@@ -1,0 +1,47 @@
+"""Plugin which add raw support.
+
+Plugin make temp directory which is use for generate jpeg from raw file (jpeg is camera-generated thumbnail).
+For this dcraw program is used.
+
+"""
+
+import logging
+
+from sigal import signals
+
+import os
+import sys
+import re
+
+from ..gallery import Image
+from ..settings import Status, get_thumb
+import copy
+
+logger = logging.getLogger(__name__)
+
+tempDir="/tmp/rawTempSigal"
+
+def pouTest(gallery):
+
+	if gallery.src_ext == '.orf':
+		logger.info('Processing RAW: '+gallery.dst_filename)
+		newFilename=re.sub('ORF', 'jpeg', gallery.dst_filename)
+		tmpname = tempDir + '/'+newFilename
+
+		os.system('dcraw -c -e "' + gallery.src_path + '" > "' + tmpname + '"')
+		
+		gallery.src_path=tmpname
+		gallery.basename = os.path.splitext(newFilename)[0]
+		gallery.dst_filename = newFilename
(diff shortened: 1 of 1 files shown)


src/sigal/plugins/lrp.py
```diff
@@ -0,0 +1,47 @@
...
+import sys
+import re
+
+from ..gallery import Image
+from ..settings import Status, get_thumb
+import copy
+
+logger = logging.getLogger(__name__)
+
+tempDir="/tmp/rawTempSigal"
```
KaibutsuX, line 22: This isn't Windows compatible

src/sigal/plugins/lrp.py
```diff
@@ -0,0 +1,47 @@
...
+		gallery.dst_filename = newFilename
+		gallery.src_filename = newFilename
+		gallery.src_ext = os.path.splitext(newFilename)[1].lower()
+		gallery.src_path = tmpname
+		gallery.thumb_name = get_thumb(gallery.settings, gallery.dst_filename)
+
+
+def register(settings):
+    logger.info('INIT lite raw plugin')
+    os.system("rm -rf "+tempDir)
```
KaibutsuX, line 44: Not Windows compatible

src/sigal/plugins/lrp.py
```diff
@@ -0,0 +1,47 @@
...
+tempDir="/tmp/rawTempSigal"
+
+def pouTest(gallery):
+
+	if gallery.src_ext == '.orf':
+		logger.info('Processing RAW: '+gallery.dst_filename)
+		newFilename=re.sub('ORF', 'jpeg', gallery.dst_filename)
+		tmpname = tempDir + '/'+newFilename
+
+		os.system('dcraw -c -e "' + gallery.src_path + '" > "' + tmpname + '"')
```
KaibutsuX, line 31: If you use subprocess you can check return codes, this is just a run it and hope it succeeds.

153957: It is also not a safe way to provide arguments to a program, provide each part as an element of a list.

>  Providing a sequence of arguments is generally preferred, as it allows the module to take care of any required escaping and quoting of arguments (e.g. to permit spaces in file names).

src/sigal/plugins/lrp.py
153957: Use `tempfile.gettempdir` instead: https://docs.python.org/3.12/library/tempfile.html#tempfile.gettempdir

saimn:
Seems I missed that PR, sorry. As discussed in #414 and #413, this should be done with the [rawpy](https://github.com/letmaik/rawpy) Python package (which is based on libraw).

PR #485, opened 2022-12-31
Added support for black detection in video thumbnail generation

KaibutsuX:
Added 3 new conf options:
* thumb_video_black_retries = 0
* thumb_video_black_retry_offset = 1
* thumb_video_black_max_colors = 4

They control how many retries a thumbnail generation should run before failure. Each retry seeks the video another N seconds determined by retry_offset in addition to the original thumb_video_delay setting.

A 'black' thumbnail is detected when the resulting image has max_colors or less distinct color values in it. This is not exclusive to black, but attempts to avoid any solid color images.


KaibutsuX:
I have no idea why my build is failing in RTD

src/sigal/video.py
```diff
@@ -228,6 +245,9 @@ def process_video(media):
                 fit=settings["thumb_fit"],
                 options=settings["jpg_options"],
                 converter=settings["video_converter"],
+                black_retries=s['thumb_video_black_retries'],
+                black_offset=s['thumb_video_black_retry_offset'],
+                black_max_colors=s['thumb_video_black_max_colors']
```
saimn, line 250: `s` should be `settings` here.

saimn (commented):
Thanks for the PR, looks good except a minor fix (see below).
I fixed the docs issue on main.

saimn:
Thanks @KaibutsuX !

PR #492, opened 2023-06-03
Fix issue with tox-gh-actions

saimn:
py310-pillow-latest: tox-gh-actions couldn't understand the parallel
option. ignoring the given option: 0


paddyroddy:
I've been wondering what was causing that error myself. Thanks!

Issue #500, opened 2023-09-30
2.4: Themes are not installed, leading to failing tests


dvzrv:
Hi! I'm packaging this project for Arch Linux. This is the build script we are using for 2.3 can be found here: https://gitlab.archlinux.org/archlinux/packaging/packages/sigal/-/blob/1f8a26e7b63f488b70085c6e23e73933cb9162a3/PKGBUILD
The current one does not apply the now irrelevant patch any longer and builds from a github provided source tarball.
Additionally, for the tests  I am now installing sigal to a temporary location to test against (more below).

When trying to build 2.4 I noticed that themes directories are no longer installed, which has the tests fail:
```
E               Exception: Impossible to find the theme /build/sigal/src/sigal-2.4/test_dir/usr/lib/python3.11/site-packages/sigal/themes/colorbox
```

Full build and test logs:
[sigal-2.4-1-x86_64-build.log](https://github.com/saimn/sigal/files/12774367/sigal-2.4-1-x86_64-build.log)
[sigal-2.4-1-x86_64-check.log](https://github.com/saimn/sigal/files/12774366/sigal-2.4-1-x86_64-check.log)

The directory structure created by installing sigal is:
```
/var/lib/archbuild/extra-testing-x86_64/dave-5/build/sigal/src/sigal-2.4/test_dir
└── usr
    ├── bin
    │   └── sigal
    └── lib
        └── python3.11
            └── site-packages
                ├── sigal
                │   ├── __init__.py
                │   ├── __main__.py
                │   ├── __pycache__
                │   │   ├── __init__.cpython-311.opt-1.pyc
                │   │   ├── __init__.cpython-311.pyc
                │   │   ├── __main__.cpython-311.opt-1.pyc
                │   │   ├── __main__.cpython-311.pyc
                │   │   ├── gallery.cpython-311.opt-1.pyc
                │   │   ├── gallery.cpython-311.pyc
                │   │   ├── image.cpython-311.opt-1.pyc
                │   │   ├── image.cpython-311.pyc
                │   │   ├── log.cpython-311.opt-1.pyc
                │   │   ├── log.cpython-311.pyc
                │   │   ├── settings.cpython-311.opt-1.pyc
                │   │   ├── settings.cpython-311.pyc
                │   │   ├── signals.cpython-311.opt-1.pyc
                │   │   ├── signals.cpython-311.pyc
                │   │   ├── utils.cpython-311.opt-1.pyc
                │   │   ├── utils.cpython-311.pyc
[...]

PR #503, opened 2023-10-27
Add data files for themes and plugins using setuptools package-data

dvzrv:
Use setuptools' `package-data` [1] to unconditionally add all files
below sigal's `themes/` dir and all data files below the
`plugins/encrypt/static/` dir to a wheel.
This works without any SCM integration and without `MANIFEST.in`.

Fixes https://github.com/saimn/sigal/issues/500

[1] https://setuptools.pypa.io/en/latest/userguide/datafiles.html#package-data

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,12 @@ license-files = ["LICENSE"]
 where = ["src"]
 namespaces = true
 
+[tool.setuptools.package-data]
+sigal = [
+    "plugins/encrypt/static/*",
+    "themes/**",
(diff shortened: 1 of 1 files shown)


saimn:
Thanks for the PR!
With `include-package-data = true` this shouldn't be needed, at least from what I understand from https://setuptools.pypa.io/en/latest/userguide/datafiles.html#subdirectory-for-data-files, but package data has always been a confusing thing with setuptools so fine by me :)

PR #506, opened 2024-01-11
Cache input_size to make updating faster

dsschult:
By adding caching for `input_size`, image files that have previously been processed do not need to be opened at all when a gallery is updated.  This improves update speed significantly, especially on slower systems with many image files.  I've seen anywhere from a 1.5x to 10x improvement with ~50k images when testing this PR on various systems.


saimn:
Thanks!

PR #507, opened 2024-01-12
Allow sorting by negative numbers

dsschult:
One of my users noticed that negative numbers specified in meta ordering weren't being sorted correctly.  This adds `ns.SIGNED` to the sorting, as well as modifying the tests to show it works.

### DIFF
--- a/tests/sample/pictures/dir1/test2/index.md
+++ b/tests/sample/pictures/dir1/test2/index.md
@@ -1,1 +1,2 @@
-Order: 01
+Order: -10
+PartialOrder: 01
--- a/src/sigal/gallery.py
+++ b/src/sigal/gallery.py
@@ -530,11 +530,11 @@ class Album:
             elif medias_sort_attr.startswith("meta."):
                 meta_key = medias_sort_attr.split(".", 1)[1]
                 key = natsort_keygen(
-                    key=lambda s: s.meta.get(meta_key, [""])[0], alg=ns.LOCALE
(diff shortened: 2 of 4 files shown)


tests/sample/pictures/dir1/test2/index.md
```diff
@@ -1 +1,2 @@
-Order: 01
+Order: -10
+PartialOrder: 01
```
saimn, line 2: The PartialOrder key doesn't seem to be used ?

saimn (commented):
Thanks for the patch, just a little question.

tests/sample/pictures/dir1/test2/index.md
dsschult: If you don't set that, then the PartialOrder test fails:

```
        settings["albums_sort_reverse"] = False
        a = Album("dir1", settings, album["subdirs"], album["medias"], gal)
        a.sort_subdirs(["meta.partialorder", "meta.order"])
>       assert [d.name for d in a.albums] == list(["test1", "test2", "test3"])
E       AssertionError: assert ['test2', 'test1', 'test3'] == ['test1', 'test2', 'test3']
E         At index 0 diff: 'test2' != 'test1'
E         Use -v to get more diff

tests/test_gallery.py:251: AssertionError
```

I suppose I could have reordered the test assertion, but it's a good check that it's using PartialOrder instead of Order.

saimn: Oh right, I missed that ! 
Sounds good them, thanks again.

PR #510, opened 2024-02-20
Only copy the static/theme files once

dsschult:
The static/theme files were being copied multiple times, every time a Writer was instantiated.  This PR makes only a single instance copy the files.


saimn:
Nice finding, thanks!
Maybe instead of using an argument to trigger the copy, this could go in a method that would be called on the first instantiation ?

dsschult:
I moved it to a method, but it needs to be an explicit call from the gallery build step or several tests fail (because only the first test instance will copy the files, and all others would not).

saimn:
Thanks!

PR #516, opened 2024-04-30
sigal serve --browser : open in your default browser

Lucas-C:
Hi!

First, thank you very much for maintaining this tool.
Thanks yo you, yesterday I was able to very quickly setup a web gallery that perfectly suited my need.

This very short contribution introduces a hand `--browser` / `-b` flag to the `sigal serve` command, that opens the sigal gallery in the user default web brower, using the standard Python module [webbrowser](https://docs.python.org/3/library/webbrowser.html).

I did not add a unit test to `tests/test_cli.py`, because it seemed difficult to me to test the standard behaviour of the `webbrowser` stanbdard package.
If need be, I can add one that ensure that `webbrowser.open()` gets called, using a mock.

### DIFF
--- a/src/sigal/__main__.py
+++ b/src/sigal/__main__.py
@@ -25,6 +25,7 @@ import pathlib
 import socketserver
 import sys
 import time
+import webbrowser
 from http import server
 
 import click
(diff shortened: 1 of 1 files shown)


saimn:
Thanks ! I had this idea a long time ago but seems it was forgotten :)
A mock test doesn't seem necessary for 2 lines.

Lucas-C:
> Thanks ! I had this idea a long time ago but seems it was forgotten :) A mock test doesn't seem necessary for 2 lines.

Thanks for handling my contribution so quickly 🙂

PR #519, opened 2024-05-26
Add basic support for .heic files

ususdei:


src/sigal/image.py
```diff
@@ -220,7 +225,11 @@ def get_exif_data(filename):
 
     try:
         with warnings.catch_warnings(record=True) as caught_warnings:
-            exif = img._getexif() or {}
+            exif = {}
+            if hasattr(img, "_getexif") and callable(img._getexif):
+                exif = img._getexif()
+            elif "exif" in img.info:
+                exif = img.getexif()._get_merged_dict()
```
saimn, line 232: `img._getexif()` is an alias for `img.getexif()._get_merged_dict()` so we can probably switch to the public methods, which I think are supported since a long time ? (Our min version of pillow is 8.0 currently)

saimn (commented):
Thanks for the PR and sorry about the delay. The PR looks good but it will need a rebase. Also would be great if you can add a test and test image (with an open source license).

src/sigal/gallery.py
```diff
@@ -264,7 +264,7 @@ def exif(self):
         datetime_format = self.settings["datetime_format"]
         return (
             get_exif_tags(self.raw_exif, datetime_format=datetime_format)
-            if self.raw_exif and self.src_ext in (".jpg", ".jpeg")
+            if self.raw_exif and self.src_ext in (".jpg", ".jpeg", ".heic")
```
saimn, line 267: The extension list here and below should also include ".heif" I guess ?
Also, those extensions should be recognized only if pillow-heif is available, so might be better to store this list somewhere where it can be imported (in the `image` module ?) and reused.

src/sigal/image.py
ususdei: Gladly!
I just wasn't sure if I would break any weird combination, because both variants use some private API.
But let tox find out :-)

src/sigal/gallery.py
ususdei: I'm not sure about `.heif` files. As I understand it, this is more of a container format and may contain different image-formats.
I guess `pillow-heif` should be able to handle them all, but I have no way to test that yet.

I would actually prefer it if the program errors out if it encounters `.heic` images without `pillow-heif` installed, instead of just silently ignoring and skipping those pictures.
I mean, they are in fact Images - no matter if supported or not - and the user probably intended for them to be included in the gallery.

saimn:
@ususdei - I played a bit with the code and activated .heic extensions by default so sigal finds them and logs a warning if pillow-heif is missing. Is it ok for you ? And just to be sure, test image is one of yours that you accept to share under sigal's license ?

ususdei:
@saimn  Yeah, sure! I'm totally okay with that. And thanks for the cleanup.

I took that random picture specifically for that purpose.
So just to give that explicitly in writing: I am the author of the photo in `tests/sample/pictures/dir1/test1/outdoor.heic` and hereby release it into public domain or alternatively under the MIT license.

saimn:
Great, thanks a lot !

PR #520, opened 2024-06-01
Implement automatic slideshows in galleria theme

ususdei:
This PR first adds the `autoplay` setting, which is just forwarded to galleria.

In then adds the [NoSleep.js](https://github.com/richtr/NoSleep.js/) library
and the setting `nosleep` to enable it.
This can prevent e.g. tablets from falling asleep during a slideshow
(note that this is not always supported by the device).


saimn:
Thanks for the PR, and sorry for the delay. I'm not about the addition of `nosleep.js`, I understand it can be useful but that's another library to manage/update, and it's easy enough to tune the OS settings if needed.

saimn:
I merged the autoplay commit, thanks.

PR #524, opened 2024-09-03
Added support for display_timestamp setting

KaibutsuX:
Shows the generation timestamp in the resulting galleries and albums

This is helpful for automated/cron updates to see when the last run was


saimn:
Thanks!

PR #525, opened 2024-10-22
Localize images from Markdown metadata

stasinos:
Read Lat, Lon fields in the Markdown files to localize images. These override the EXIF GPS coordinates, if both are present, allowing manually localizing the images on the map. This is useful to, for instance, localize the image based on what is depicted instead of where the camera was standing.

### DIFF
--- a/docs/image_information.rst
+++ b/docs/image_information.rst
@@ -16,6 +16,10 @@ keys are used by Sigal to get the useful informations on the gallery:
 - *Title*: the image title.
 - *Date*: the file date, useful when it cannot be read from the EXIF metadata,
   e.g. for videos and some image formats.
+- *Lat* and *Lon*: geo-location for positioning the image on map galleries.
+  Note that this overrides EXIF coordinates, so except for localizing when
+  EXIF cooredinates are missing it can also be used to localize the image
+  based on what is depicted instead of where the camera was standing.
 
 Any additional meta-data is available in the templates. For instance::
 
--- a/docs/image_information.rst
+++ b/docs/image_information.rst
@@ -18,7 +18,7 @@ keys are used by Sigal to get the useful informations on the gallery:
   e.g. for videos and some image formats.
 - *Lat* and *Lon*: geo-location for positioning the image on map galleries.
   Note that this overrides EXIF coordinates, so except for localizing when
-  EXIF cooredinates are missing it can also be used to localize the image
+  EXIF coordinates are missing it can also be used to localize the image
   based on what is depicted instead of where the camera was standing.
 
 Any additional meta-data is available in the templates. For instance::
--- a/src/sigal/gallery.py
+++ b/src/sigal/gallery.py
@@ -5,6 +5,7 @@
 # Copyright (c) 2017      - Mate Lakat
 # Copyright (c) 2018      - Edwin Steele
 # Copyright (c) 2021      - Tim AtLee
+# Copyright (c) 2024      - Stasinos Konstantopoulos
 
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to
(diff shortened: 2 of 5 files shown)


docs/image_information.rst
```diff
@@ -16,6 +16,10 @@ keys are used by Sigal to get the useful informations on the gallery:
 - *Title*: the image title.
 - *Date*: the file date, useful when it cannot be read from the EXIF metadata,
   e.g. for videos and some image formats.
+- *Lat* and *Lon*: geo-location for positioning the image on map galleries.
+  Note that this overrides EXIF coordinates, so except for localizing when
+  EXIF cooredinates are missing it can also be used to localize the image
```
saimn, line 21:
```suggestion
  EXIF coordinates are missing it can also be used to localize the image
```

src/sigal/gallery.py
```diff
@@ -286,6 +287,15 @@ def _get_markdown_metadata(self):
 
         return meta
 
+    @cached_property
+    def lat(self):
+        return self.markdown_metadata.get("lat", {})
```
saimn, line 292:
```suggestion
        return self.markdown_metadata.get("lat", "")
```
should return a string by default I think ?

src/sigal/gallery.py
```diff
@@ -286,6 +287,15 @@ def _get_markdown_metadata(self):
...
+
+    @cached_property
+    def lon(self):
+        """Other metadata extracted from the Markdown index.md file."""
+        return self.markdown_metadata.get("lon", {})
```
saimn, line 297:
```suggestion
        return self.markdown_metadata.get("lon", "")
```

src/sigal/gallery.py
saimn, line 292: Can you change the properties to return `media.exif.gps.lat/lon` if not available in meta ? this would avoid the if/else in the template.

saimn (commented):
Seems useful, thanks. A few comments below.

src/sigal/gallery.py
stasinos: Or None which simplifies the code even further, allowing get() its default behaviour.
The related if statement in Jinja works the same.

src/sigal/gallery.py
stasinos: Or None which simplifies the code even further, allowing get() its default behaviour.
The related if statement in Jinja works the same.
Please see new commit.

src/sigal/gallery.py
stasinos: But that would make it impossible to write a template that prefers the EXIF coordinates.
You might not want this, IMHO.

saimn: people can still use directly `media.exif.gps`. But for `media.lat` I think it makes sense to avoid the manual coordinates or the exif ones.

stasinos: I hate to insist, but I feel I have to. I perceive the Python code as a considerably "deeper" layer than the templates. I expect that many users will be happy to work at the template layer but will want to avoid the Python layer. Thus, I expect that a simple, easily predictable behaviour for media.lat/lon and media.exif.gps makes it easy to write templates.

Now, if you really want to have a template without ifs, I suggest keeping meta.lat/lon as they are and adding a new set of variables that implement the fallback and a configuration option that defines preference to GPS or .md if both are present.

stasinos: And, BTW, come to think of it, one of the ifs will still be needed as it might be the case that there are no coordinates at either EXIF or .md. So we will still need an if, except without the else branch.

stasinos: I just had another idea: How about a config option that sets the order of preference between md, EXIF?
Then both ways are accessible at the template. If the option also includes a "default location" (say, 0,0 or someplace in the middle of the Pacific) then all photos are guaranteed to have a location and the template can be completely without checks. This has the advantage that photos without a location do not silently disappear but are placed at a location that the users know to mean that they have forgotten to add a location.

PR #536, opened 2025-08-11
nonmedia_files plugin: adding support for PDF thumbnails using optional dependency pdf2image

Lucas-C:
I was able to extend the `nonmedia_files` plugin to use https://github.com/Belval/pdf2image in order to generate thumbnails for PDF files.

I also add to slightly extend the `album.html` template to support this.
I only implemented it for the `colorbox` & `photoswipe` themes.

This draft PR demonstrate how it can be done, but requires a bit more work before being ready to be merged.

Would you be open to this feature?
Is this PR approach of using an optional depenency OK?

Preview screenshot:
image

PR checklist:
* [x] unit tests are passing & code coverage is ✅
* [x] documentation has been added, in `nonmedia_files.py` module docstring
* [x] I had already added my name to the `AUTHORS` file


saimn:
Sorry for the late answer. 

> Would you be open to this feature?
> Is thisis PR approach of using an optional depenency OK?

As long as it is optional and only concerns a plugin I think it's fine. I'm a bit more concerned about the impact on jinja templates, since maintenance of the themes is a bit more painful, but most of the change look more like some refactoring?

Lucas-C:
> As long as it is optional and only concerns a plugin I think it's fine.

Great!

> I'm a bit more concerned about the impact on jinja templates, since maintenance of the themes is a bit more painful, but most of the change look more like some refactoring?

Yes, I tried to simplify the code.

tox.ini
```diff
@@ -19,6 +19,8 @@ deps =
     pillow90: Pillow==9.0.1
     pillow100: Pillow==10.0.1
     pillow110: Pillow==11.0.0
+    # We only install the pdf2image package in one environement:
+    pillow-latest: pdf2image
```
saimn: can you put it in `pyproject.toml` instead ? (in optional deps)

saimn (commented):
Thanks, just a small request below.

Lucas-C:
Things should be better now:
```
$ python -m coverage run -m pytest tests/test_plugins.py && coverage report -m
...
tests/test_plugins.py ....                                                                                           [100%]

=============================================== 4 passed in 96.26s (0:01:36) ===============================================
Name                                    Stmts   Miss  Cover   Missing
---------------------------------------------------------------------
...
src/sigal/plugins/nonmedia_files.py        70      2    97%   83, 101
```

Lucas-C:
Damn, it's still not working...
I added `tests/sample/pictures/nonmedia_files/dummy.md` but there is still no code coverage on the `generate_thumbnail()` call from `src/sigal/plugins/nonmedia_files.py` in Codecov...
Any idea?

saimn:
And yes we could test only on one specific job but it's easier if we have the same testing configuration everywhere.

Lucas-C:
> `.md` files are excluded by default (probably because those are used for sigal's album & image descriptions), which is why I was suggesting `.txt` or `.csv` :)

Oh, OK!

I changed that and hooray, the pipeline is now ✅

